PR #20, opened 2019-02-08
Packaging

lasofivec:
Changes on the packaging of the library. Mostly on how to choose which version of the 'polygon' library to choose.


lasofivec:
Seems like the tests are failing because the branch devel was also broken...

Didou09:
Correct, the tests are not updated yet so approximately half of them are failing on branch devel, not your fault ;-)

conda_recipe/meta.yaml
```diff
@@ -40,9 +46,14 @@ requirements:
...
     - pygments
     - pandas
-    - {{ environ['PCK'] }}
+    - polygon2 # [py27]
+    - polygon3 # [not py27]
 
 test:
+  requires:
+    - nose
+    - polygon2 # [py2k]
```
Didou09: py27 instead of py2k ?

Didou09: Please change to py27 unless good reason to leave py2k (explain ?)

Didou09:
Please address the changes requested, then I will validate and merge :-)

lasofivec:
It seems you corrected the changes (ie. "py2k" -> "py27"). So the issue is solved ?

For future reference, `py2k` is for any python of major version = 2 (which is what is needed by `polygon2`), since ToFu only supports python2.7 of the 2.x versions, for us `py2k` and `py27` are equivalent

https://docs.conda.io/projects/conda-build/en/latest/source/define-metadata.html?highlight=meta.yaml#preprocess-selectors

PR #22, opened 2019-02-08
Original LM's branch, cleaning changes and more

lasofivec:


lasofivec:
The branch `laura` was mostly minor changes not important but no to loose them I still merges with `LOS_testing`...

lasofivec:
@Didou09 I think you created the `laura` branch, so it is protected. 
Now that it's merged, would you minds erasing it ?

PR #24, opened 2019-02-19
Los testing

lasofivec:
Optimization on LOS ray-tracing, plus minor packaging updates


Didou09:
Sounds good, let's try it

lasofivec:
I restored the branch, will keep it for a couple of weeks just in case 👍

PR #28, opened 2019-03-11
Los testing

lasofivec:
update for memory optimization
changes on kmin-kmax algo


lasofivec:
The plot_touch is not working (on devel, thus also not in this commit).
And there is an issue on 2.7 that is on devel as well.

PR #97, opened 2019-06-14
Vignetting

### LINKED ISSUE #70: LOS vignetting : Profile / optimize / parallelize
Profile, optimize and parallelize the best algorithm

Make the structure of the code (inlined basic functions...) allows for creating other version later if necessary

Make available through a single python function, for N LOS and M polygons, return a (N,M) array of booleans

lasofivec:
## Vignetting

- Added functions to compute the 3d bounding box of a polygon in 3d space
- Added functions for the triangulation of a polygon by earclipping
- Added function that tests if intersection between ray and triangle in space
- Added functions for vignetting : for a set of rays and a sets of polygons, test if the rays go through the polygons or not.

## Bugs

- Found several small bugs in the function that computes for a list of flux surfaces and a list of rays the kmin, kmax

## Unit test

- Added unit tests for triangulations, and vignetting
- Added unit test for computation of kmin,kmax

## Installation

- Had to use some cpp tools, so had to update the `setup.py`
- Updated the `setup.py` to be able to "clean" an installation (erase compilation files)
- Took out the possibility of not using cython, as this is now impossible

## Issues
 
Fixes #97 
Fixes #70


setup.py
```diff
@@ -281,6 +332,7 @@ def check_for_openmp(cc_var):
     #},
 
     ext_modules = extensions,
-    #cmdclass={'build_ext':build_ext},
+    cmdclass={'build_ext':build_ext,
+              'clean':CleanCommand},
```
lasofivec: This adds the possibility to do
`python setup.py clean`
and that it actually cleans something...

Didou09:
No accepting as such because on test [failed](https://travis-ci.org/ToFuProject/tofu/builds/545814977):

Please correct...

But thanks a lot for the hard work and the detailed summary of the changes ! :-)

Issue #99, opened 2019-06-18
polygon sampling failing 


Didou09:
On branch Issue61_IMAS2TofumoreRobust, the polygon sampling routines seem to be failing

lasofivec:
What is the error ?
Could it be the same as Issue #84 ?

Didou09:
olved by PR #127

PR #100, opened 2019-06-18
Issue61 imas2tofu more robust

### LINKED ISSUE #57: Finish and test saving to/ loading from .mat files
Support for matlab users

### LINKED ISSUE #61: Make imas2tofu more robust
versus:
* geom=True and data=False (define chans) in load_Diag()
* user providing ids or dids or (shot, user....) => implement methods in objects using utils
* use same names of diagnostics as IMAS

### LINKED ISSUE #65: In tofu 1.4.0: plot_combine() mixing up where to plot geom
Reproduce with (on WEST):

```python
import tofu as tf

dids = dict(shot=53069)
interf = tf.imas2tofu.load_Diag(diag='interfero', dids=dids, Name='test')
bolo = tf.imas2tofu.load_Diag(diag='bolo', dids=dids, geom=False, Name='test')
ece = tf.imas2tofu.load_Diag(diag='ece', dids=dids, geom=False, Name='Test')

ece.plot_combine([interf, bolo])
```

### LINKED ISSUE #67: Make installing with easyconfig easier for ITER clusters
The dependency on git during setup is an issue wen installing with easyconfig:
see https://jira.iter.org/browse/IMAS-2287?filter=-3 for context

**_Explanation:_**

> I got your email Vezinet Didier, asking how to make the package easier (i.e. getting rid of the patch).
> 
> What the patch does is remove the part that requires git commands during the python setup.py build step. Why? because the .eb file downloads the tarball – so there is no git repository inside.
> 
> How the packaging would be better: remove the need for git. You can probably include a test that inspects whether setup.py is run from a git repo or not. E.g. check for the .git directory's existence, or make git describe fail gracefully and continue with the default version – now it breaks

### LINKED ISSUE #94: Operator overloading for CamLOS1D
Would be nice to have an easy way of concatenating CamLOS1D
=> overload __add__

Didou09:
This PR addresses mostly imas interafcing, as well as a few other issues.
It does not close the branch because there are still a few things to implement, but this PR allows to update devel before branching from it to address solid angles.

Main changes
=========
IMAS interfacing
------------------

1. **MultiIDSLoader** class now handles all imas interfacing:
- allows for loading several ids from different idd
- dynamic printing of ids loading
- get_summary() methods giving an overview of content
- provides a dictionary of shortcuts for fast typing of pre-recorded imas signals
- provides dictionary of preset ids and signals
- Flexible instanciation, from idd of idd args, ids or ids args, open or not, get or not
- provides add_idd(), remove_idd(), add_ids(), remove_ids()
- Include to_Config(), to_Plasma2D(), to_Diag() to export to tofu objects

2. **All stored in a single _core.py**:
- MultiIDSLoader 
- load_Config(), load_Plasma2D(), load_Diag()

3. **save_to_imas()**
- Actual routines implemented in tf.imas2tofu._core
- methods implemented for Struct, Config, CamLOS1D

Standardization:
------------------
- get_summary() is now a method of the AbstractToFuObject parent class, implemented for Config, CamLOS1D, Plasma2D, 

cameras defining:
--------------------
- There was a bug in some special cases of 1D camera definition, when all Ds are on a similar plane sharing the same D for the 2 first LOS, solved
- overloaded __add__ operator to allow for easy CamLOS1D concatenation
- 

Visualization:
--------------
- There was a bug in tf.data._plot_combine() (wrong graph), solved (issue #65 )
- There was another bug with plot_combine() when several equilibria were provided, the reference time was not properly defined (short-term fix, on the long term issue #79 should fix it)

saving/loading tofu object
-----------------------------
- saving to .mat file is now more robust (operational)

setup.py:
----------
- git dependency is now optional (issue #67 )

left TODO:
=======
Features:
----------
- Finish implementing save_to_imas() for DataCam1D
- Implement load_from_imas()
- Implement Rays.calc_signal_from_Plasma2D()
- Implement bash tofuplot ?

Known issues:
---------------
- issue with get_sampleEdge() and get_sample_S() (issue #99)

Fixes issues:
========
Fixes #57 (saving / loading to/from mat files)
Fixes #65 (plot_combine() plotting LOS in wrong axes)
Fixes #67  (git dependency)
Fixes #94 (operator overloading for CamLOS1D)

In the end (after additional features added), will fix #61 (imas2tofu more robust)


.gitignore
```diff
@@ -53,4 +54,4 @@ Details
 usr/
 
 # tests of LOS on west and ITER
-tofu/tests/tests01_geom/laura_tmp/
\ No newline at end of file
+tofu/tests/tests01_geom/laura_tmp/
```
lasofivec: I think we can erase this line ! ;)

tofu/geom/_core.py
```diff
@@ -31,6 +31,7 @@
 
 
 # ToFu-specific
+from tofu import __version__ as __version__
```
lasofivec: is the `as ...` really necessary ?

PR #107, opened 2019-07-05
Issue71 solid angle particle vectorized

Didou09:
see PR #104  and PR #106


lasofivec:
I think @Didou09, as author of this PR, you should be able to update the branch `Issue71...` to its latest version... 
If not we can close the pull request (without merging) and make a new one with latest version of branch (See #111 )

PR #118, opened 2019-07-25
Issue108 calc los signal

### LINKED ISSUE #108: calc_LOS_signal2 : choice of strategy and mutualization of sampling
In _GG:

- LOS_calc_signal() should be replaced by LOS_calc_signal2()
- LOS_calc_signal2() has been drafted (branch from Issue61_Imas2TofuMoreRobust)

- It introduces: a choice between three strategies (minimize user's function calls by using vectorized routines for small number of LOS, minimize memory usage by looping for large number of LOS)

- It needs both the existing LOS_get_sample (for minimize = calls) and a single-LOS version of sampling (for minimize = memory)

- Making that single-LOS version of LOS_get_sample() may require revizing a bit the _sampling_tools.pxd to introduce inlined single-LOS unparallelized versions of all sampling methods.

This update will make computation of synthetic signal faster / more efficient (especially for large number of LOS) and more robust vs user's functions (if ill-defined, we can minimize the number of calls).

It will also make synthetic signal computation from Plasma2D objects easier and more natural

lasofivec:
This answers Issue #108 

> In _GG:
> 
> - LOS_calc_signal() should be replaced by LOS_calc_signal2()
> - LOS_calc_signal2() has been drafted (branch from Issue61_Imas2TofuMoreRobust)
> 
> - It introduces: a choice between three strategies (minimize user's function calls by using vectorized routines for small number of LOS, minimize memory usage by looping for large number of LOS)
> 
> - It needs both the existing LOS_get_sample (for minimize = calls) and a single-LOS version of sampling (for minimize = memory)
> - Making that single-LOS version of LOS_get_sample() may require revizing a bit the _sampling_tools.pxd to introduce inlined single-LOS unparallelized versions of all sampling methods.
> 
> This update will make computation of synthetic signal faster / more efficient (especially for large number of LOS) and more robust vs user's functions (if ill-defined, we can minimize the number of calls).
> 

It will also make synthetic signal computation from Plasma2D objects easier and more natural
**Done** the single-LOS version had to be done for each quadrature rule and I took out the python interface for LOS_get_sample_single, as I don't think this is useful. I just implemented a version in inlined cython.

For more comments see the *code review*.

**PS:** this branch is not passing travis tests. This however is beyond my reach since the errors are coming from `Issue61_Imas2TofuMoreRobust`

Fixes #108


tofu/geom/_GG02.pyx
```diff
@@ -2620,86 +2620,17 @@ def vignetting(double[:, ::1] ray_orig,
...
-        # else:
-            # _st.romb_left_rule_abs(num_los, val_resol,
-                                   # &DLs[0,0], &DLs[1, 0],
-                                   # &dLr[0], &los_coeffs, &los_ind[0],
-                                   # num_threads=num_threads)
-    pass
-
-
-# TO BE USED IN LOS_calc_signal2 FOR MEMORY MINIMIZATION ?
-cdef inline np.ndarray LOS_get_sample_single(D, u, resol, method, mode):
```
lasofivec: This two functions were redundant and all inline and totally cythonized/C, so I wrote it in `_sampling_tools.py`

tofu/geom/_GG02.pyx
```diff
@@ -2620,86 +2620,17 @@ def vignetting(double[:, ::1] ray_orig,
...
-    # and return k array + scalar effective resolution
-
-
-
-#
-#       TO BE MODIFIED TO RELY ON INLINED single-LOS routine _st.all_rules ?
-#
-def LOS_get_sample(double[:,::1] ray_orig, double[:,::1] us, dL,
-                   double[:,::1] DLs, str dmethod='abs',
+def LOS_get_sample(int num_los, dL, double[:,::1] DLs, str dmethod='abs',
```
lasofivec: I realized `ray_orig` and `ray_vdir` (aka `Ds` and `us`) are not used in the function, so took them out...

lasofivec: Only `num_los` is actually needed

tofu/geom/_GG02.pyx
```diff
@@ -2740,26 +2667,19 @@ def LOS_get_sample(double[:,::1] ray_orig, double[:,::1] us, dL,
...
         assert sz1_dls == 2, "Dim 0 of arg DLs should be 2"
-        error_message = "Args ray_orig, us, DLs should have same dimension 1"
-        assert sz2_ds == sz2_us == sz2_dls, error_message
+        error_message = "Args DLs should have dim 1 = num_los"
+        assert num_los == sz2_dls, error_message
         bool1 = not dl_is_list and dL > 0.
-        bool2 = dl_is_list and len(dL)==sz2_ds and np.all(dL>0.)
-        assert bool1 or bool2, "Arg dL must be a double or a List, and all dL >0.!"
+        bool2 = dl_is_list and len(dL)==num_los and np.all(dL>0.)
+        assert bool1 or bool2, "Arg dL must be a double or a List, and dL >0.!"
```
lasofivec: lighter tests since less parameters...

tofu/geom/_GG02.pyx
```diff
@@ -2957,302 +2876,13 @@ def integrate1d(y, double dx, t=None, str method='sum'):
...
-                     str method='sum', bint ani=False,
-                     t=None, fkwdargs={}, str minimize='calls',
-                     bint Test=True, int num_threads=16):
+def LOS_calc_signal(func, double[:,::1] Ds, double[:,::1] us, dL,
+                    double[:,::1] DLs, str dmethod='abs',
+                    str method='sum', bint ani=False,
+                    t=None, fkwdargs={}, str minimize='calls',
+                    bint Test=True, int num_threads=16):
     """ Compute the synthetic signal, minimizing either function calls or memory
+    TODO: since we are working in cython... the least is to give "func" 's signature
```
lasofivec: As said in the comments: the signature of the function `func` should be given

tofu/geom/_GG02.pyx
```diff
@@ -3306,22 +2935,28 @@ def LOS_calc_signal2(func, double[:,::1] Ds, double[:,::1] us, dL,
...
     # Minimize function calls: sample (vect), call (once) and integrate
-    if minimize == 'calls':
+    if minim == 'calls':
         # Discretize all LOS
-        k, reseff, ind = LOS_get_sample(Ds, us, dL, DLs,
-                                        dmethod=resMode, method=method,
+        k, reseff, ind = LOS_get_sample(num_los, dL, DLs,
+                                        dmethod=dmode, method=imode,
                                         num_threads=num_threads, Test=Test)
-        nbrep = np.r_[ind[0], np.diff(ind), k.size - ind[-1]]
```
lasofivec: no negative indexing !!

tofu/geom/_GG02.pyx
```diff
@@ -3330,52 +2965,461 @@ def LOS_calc_signal2(func, double[:,::1] Ds, double[:,::1] us, dL,
...
         # Integrate
         if method=='sum':
-            for ii in range(0,ND):
-                sig[:,ii] = np.sum(val[:,indbis[ii]:indbis[ii+1]], axis=-1)*reseff[ii]
+            for ii in range(num_los):
+                sig[:,ii] = np.sum(val[:,indbis[ii]:indbis[ii+1]],
+                                   axis=-1)*reseff[ii]
         elif method=='simps':
-            for ii in range(0,ND):
+            for ii in range(num_los):
```
lasofivec: I found that when `range(0,n)` is used instead of `range(n)` cython actually first creates a tuple for the `(0,n)` that is passed later to the range function. Seems like a silly little conversion, so I just avoid indicating the `0`

tofu/geom/_GG02.pyx
```diff
@@ -3330,52 +2965,461 @@ def LOS_calc_signal2(func, double[:,::1] Ds, double[:,::1] us, dL,
...
+                    dsbis = np.repeat(Ds[:,ii].reshape((3,1)), sz_coeff, axis=1)
+                    usbis = np.repeat(us[:,ii].reshape((3,1)), sz_coeff, axis=1)
+                    ksbis = np.asarray(<double[:sz_coeff]>los_coeffs[0])
+                    # loop over time for calling and integrating
+                    for jj in range(nt):
+                        val = func(dsbis + ksbis * usbis,
+                                   t=ltime[jj], **fkwdargs)
+                        sig[jj, ii] = scpintg.romb(val, show=False,
+                                                   dx=loc_eff_res[0])
+    # --------------------------------------------------------------------------
```
lasofivec: I know all these different cases seem heavy, but I believe they are necessary to avoid a `if` statement inside a loop-loop.

tofu/geom/_core.py
```diff
@@ -4286,15 +4286,16 @@ def get_sample(self, res=None, resMode='abs', DL=None, method='sum', ind=None,
 
         # Launch    # NB : find a way to exclude cases with DL[0,:]>=DL[1,:] !!
         # Todo : reverse in _GG : make compact default for faster computation !
-        k, reseff, ind = _GG.LOS_get_sample(Ds, us, res, DL,
+        nlos = Ds.shape[1]
+        k, reseff, lind = _GG.LOS_get_sample(nlos, res, DL,
```
lasofivec: I saw the variable name `ind` was being used twice ! renamed it to `lind` for that reason

tofu/geom/_core.py
```diff
@@ -4576,9 +4577,10 @@ def calc_signal(self, ff, t=None, ani=None, fkwdargs={}, Brightness=True,
         # Exclude Rays not seeing the plasma
         s = _GG.LOS_calc_signal(ff, Ds, us, res, DL,
                                 dmethod=resMode, method=method,
-                                t=t, Ani=ani, fkwdargs=fkwdargs, Test=True)
+                                t=t, ani=ani, fkwdargs=fkwdargs, Test=True)
```
lasofivec: For the two changes above:
It seems to me more are and more obvious that we need to make a convention for naming variables. Mostly when to use capital letters or not. The branch Issue61 was not working because there were many instances where `DL` was being used instead of `Dl`, and in other functions we have the variable `DL` and `dL`. Keyword arguments also change from one function to the next one, eg. some times we have `func_a(..., test=True)` then `func_b(..., Test=True)`
Not only the difference in the capitalization gives no clue as to what is that variable, but it leads to confusion.
More explicit variable names would also imply less need for long comments/documentation.

tofu/geom/_core.py
```diff
@@ -4576,9 +4577,10 @@ def calc_signal(self, ff, t=None, ani=None, fkwdargs={}, Brightness=True,
...
         if t is None or len(t)==1:
             sig[indok] = s
+            sig = sig.reshape((1,len(sig)))
```
lasofivec: This might not be needed, I added it wanting to make it the array 2D when creating the 2D camera in post treatment, however when the camera is created, it "squeezes" the array, and flattens it again

tofu/geom/utils.py
```diff
@@ -693,7 +693,7 @@ def create_config(case=None, Exp='Dummy', Type='Tor',
                   Lim=None, Bump_posextent=[np.pi/4., np.pi/4],
                   R=2.4, r=1., elong=0., Dshape=0.,
                   divlow=True, divup=True, nP=200,
-                  out='object', SavePath='./'):
+                  out='object', SavePath='./', path=_path_testcases):
     """ Create easily a tofu.geom.Config object
```
lasofivec: sometimes I run tests from other paths, so I added to possibility to determine the path for testing

tofu/geom/_GG02.pyx
lasofivec: func(pts, t=None, vect=None) => data 
pts : ndarray (3, n)
vect : ndarray(3, n)
t = ndarray(m)
data : ndarray(n) if t is None, else ndarray(m,n)

tofu/geom/_GG02.pyx
```diff
@@ -3330,52 +2965,461 @@ def LOS_calc_signal2(func, double[:,::1] Ds, double[:,::1] us, dL,
...
+                                                         &loc_eff_res[0],
+                                                         &los_coeffs[0])
+                    dsbis = np.repeat(Ds[:,ii].reshape((3,1)), sz_coeff, axis=1)
+                    usbis = np.repeat(us[:,ii].reshape((3,1)), sz_coeff, axis=1)
+                    ksbis = np.asarray(<double[:sz_coeff]>los_coeffs[0])
+                    val = func(dsbis + ksbis * usbis,
+                               t=ltime, vect=-usbis, **fkwdargs)
+                    # loop over time for integrating
+                    for jj in range(nt):
+                        sig[jj, ii] = scpintg.romb(val, show=False, axis=1,
```
lasofivec: no loop in time:
sig[:,ii] = integration_method(val, .. axis=1)

PR #120, opened 2019-07-29
Issue108 calc los signal

lasofivec:
Took out useless loop when using hybrid  algorithm in the `calc_signal` function 
In `geom/_core.py` fixed in-/output treatment of the signal.
None of the fails of unit tests are regarding the `calc_signal` function.

## Extra
Rename all variables `num_los` to `nlos` to be more homogeneous throughout the code.


lasofivec:
I need to move on so I'm going to accept this pull request...

Issue #138, opened 2019-08-22
LOS_calc_singal


lasofivec:
Error: cannot reshape memory view.
Arguments Ds, us, in LOS_calc_signal must be numpy arrays

Didou09:
Fixed in devel by PR #141

Didou09:
Closed in 1.4.1

Issue #137, opened 2019-08-20
calc_signal_from_Plasma2D should use new _GG.LOS_calc_signal


Didou09:
Implement the possibility of using the new version of _GG.LOS_calc_signal() for tf.geom.Rays.calc_signal_from_Plasma2D().

Use a keyword to switch between methods, benchmark all options and set the best as default.

Didou09:
Opened Issue #138  and Issue # 142 from this one

Didou09:
Fixed (devel) by PR #151

Didou09:
Closed in 1.4.1

PR #139, opened 2019-08-22
Issue137 rays calc from plasma2 d use new gglos calc signal

Didou09:
While becnhmarking (Issue #137) Found a bug in _GG.LOS_calc_signal() for minimize = 'memory' and 'hybrid'.
=> This PR is necessary to update devel with last changes so Laura can branch for new Issue #138


Didou09:
Not passing tests

Didou09:
Passing test with last commit

Issue #142, opened 2019-08-22
LOS_calc_signal computation of quadrature rules optimization


lasofivec:
When computing quadrature rules, for some with do it in two steps: the first one to compute the total size of quadrature points and the second one, after the creation of the arrays, fills the latter with the computed values.
For others with do these two steps in one loop and realloc the arrays at each step.

It might be quicker to do this always in two steps and thus be able to parallelize both loops.

Didou09:
Fixed, in deve, by PR #164

Didou09:
Closed in 1.4.1

PR #148, opened 2019-08-23
Issue137 rays calc from plasma2 d use new gglos calc signal

Didou09:
Main changes to benchmark 
(more robust and practical)
===================

info of tofu version used
---------------------------
* the benchmark now automatically prints and writes to the saved file the current tofu version number and the location of the repository where it was imported from 

Results saved on the go
--------------------------
* They are (progressively) instead of waiting until the end to save everything, this way we still get partial results if execution fails after a while

No need to define emissivity function
-----------------------------------------
* A default emissivity function is already defined in the module
* pass **func=True** to use it

Better saving file name
-------------------------
* The default name of the saving file now includes the current host name (using socket)
* The default name of the saving file now includes the current user (using getpass)

Customizable saving file name
----------------------------------
 * In addition to the default file name, a user-provided char str can be appended at the end of the name
* Useful for remembering key aspects / adding comments not already contained in the name

prints to a txt file
-------------------
* prints can now be redirected to a custom txt file instead of the console (sys.stdout)
* The txt file automatically inherits the name of the file used for saving data, the extension only differs (.txt instead of .npz)
 * pass **txtfile=True** to use it

usable from bash:
--------------------
* using argparse, the python script can be called directly from bash
* bash arguments are passed to the python function
* e.g. `--func` (bash) is equivalent to `func=True` (python)

Example:
======
To launch a benchmark from bash and write everything in files and customizing the file names with, for example, the name of the server (inria) on which the benchmark is run, just do:
```{code:bash}
python ./calc_signal_benchmark.py -txtfile True -nameappend ''testforchecking"
```

Issues:
====
Does not close Issue #137  yet but contributes to it
Useful too for Issue #142


lasofivec:
Seems to me there is a mistake in the example

`python ./calc_signal_benchmark.py --txtfile True --nameappend ''testforchecking"`

Issue #160, opened 2019-08-27
_GG.LOS_get_sample(): bug in mode=rel and slowdown


Didou09:
Minimum working example:
==================

Considering 2 LOS, we want to sample them from in the intervals k=[1,2] and [10,20] respectively. All k output should be in these intervals, but:
dL = 0.25
DL = np.array([[1.,10.],[2.,20.]])
k = _GG.LOS_get_sample(2, dL, DL, dmethod='rel', method='sum')[0]
print(k)
[80., 52.8, 576., 52.8, 11.25....]

Problem appears when:
* dL unique + rel + sum    => Error !!  (k too large)
* dL unique + rel + simps  => ok
* dL unique + rel + romb  => ok
* dL unique + abs + sum  => ok
* dL unique + abs + simps  => ok
* dL unique + abs + romb  => ok

* dL unique + rel + sum  => Error !  (k too small)
* dL unique + rel + simps  => Error ! (k too small)
* dL unique + rel + romb  => Error!   (k too small)
* dL unique + abs + sum  => ok
* dL unique + abs + simps  => ok

When k is too small, it seems to be due to line 733 in tofu/geom/_sampling_tools.pyx (for the case 'simps', but I guess it is similar for others): the local resolution should not be 1/num_raf but seg_length / num_raf like in line 782.
The 'relative' aspect is already taken into account in line 730 (vs 779).

@lasofivec, please have a look, debug and double-check all results

For the case k too large... no time to check, please investigate

Also, optimization:
============
In almost all routines in _sampling_tools.pyx, there is a if test (if ii == 0) inside a for loop.
Like in previous case, please take the test out (run first iteration separately) to speed up computation

Maintenance:
=========
* It looks like many of the lins inside the for loops have a lot (almost everything) in common, please consider inlining for easier maintenance and better readability.
* Functions like simps_left_rule_abs_single (line 750) and simps_left_rule_rel_single (line 703) seem to be exactly the same . If so, consider unifying / deleting duplicate for better readability and maintenance

lasofivec:
in which branch ?? devel ?

lasofivec:
There is a bug for ``k = _GG.LOS_get_sample(2, dL, DL, dmethod='rel', method='sum')[0]``
Actually found it in another branch that I am working on right now, but I will submit the bug fix rn.

However, for the other methods, I haven't found any issues. I added a unit test:

> import tofu.geom._GG as _GG
> import numpy as np
> 
[...]

PR #161, opened 2019-08-27
[bug-fix] in get_sample 

lasofivec:
For  ``dmethod='rel'`` with ``method='sum``, there was a bug due to a wrong indentation.
Fixed and added a unit test to make sure this issue will not be found again


Didou09:
Partially fixes Issue #160

PR #164, opened 2019-08-28
Issue142 calc signal perf

lasofivec:
### Overall
- Using ``-O3`` flag instead of ``-O0``  for faster execution time even if compilation is slower
- More use of memory views

### LOS_get_sample

- Created two inlined, cython function to sample line of sight: one for a fixed resolution for all LOS, another with variable resolutions
- Using C-arrays when possible to accelerate element access
- Took out ``if ii==0:`` statement in loops

### LOS_calc_signal

- For all methods: using more memory views and/or pointer for speed up
- For integration method ``sum`` instead of using **numpy**'s version we use (when performance is better) a cython version
- Tried different cython versions of ``np.sum`` left them for the moment, we will see in the future if any should be removed

## Unit test

- Testing all get_sample options in a short new unit test
- Testing all options for LOS_calc_signal (method of discretization, of integration, steps relative, absolute, unique or changing for each los, etc.). This is test might be too big... @Didou09 

## Styling code

- Uniformization of variables names. For example ``num_los`` and ``nlos`` now to ``nlos``
- More readable names. For example avoiding different variables ``dL, DL, Dl, dl"
- Avoided repetition of code by using inline functions
- Quadrature rules methods no longer accessible from outside


tofu/tests/tests01_geom/tests03_core.py
```diff
@@ -856,22 +856,28 @@ def ffT(Pts, t=None, vect=None):
...
-                                          res=0.01, DL=None, resMode='abs',
-                                          method='simps', minimize=mmz, ind=ind,
-                                          plot=False, out=np.ndarray,
-                                          fs=(12,6), connect=connect)
-                    sig, units = out
-                    assert not np.all(np.isnan(sig)), str(ii)
+        for aa in [True, False]:
+            for rm in ["abs", "rel"]:
+                for dm in ["simps", "romb", "sum"]:
+                    for mmz in minimize:
```
lasofivec: These loops make the test much slower but it's the most complete test we can have... toss or keep?

lasofivec:
Fixes #142

tofu/tests/tests01_geom/tests03_core.py
Didou09: Keep, and if want faster tests at some point we can temporarily comment them

PR #173, opened 2019-09-05
Euroscipy2019

Didou09:
The talk has been done, this is the final version of the presentation


Didou09:
You're right, I added compiled files by mistake yesterday,
Thanks for removing them

benchmarks/calc_signal_benchmark.py
```diff
@@ -42,9 +42,9 @@
 # Defining defaults
 ###################
 
-_LRES = [-1,-3,0]
-_LLOS = [1,3,0]
-_LT = [1,3,0]
+_LRES = [-3,-3,0]
+_LLOS = [5,5,0]
+_LT = [3,3,0]
```
lasofivec: this is probably shouldn't be part of the PR. We can leave it if you want @Didou09, but for cleaningness maybe you want to remove it ?

Didou09:
Any more changes ? I'm gonna accept the pull request

lasofivec:
You can ofc :) but I did do a comment in the code (see in this same thread above).

benchmarks/calc_signal_benchmark.py
Didou09: Hum, not sure, it's good to have easy access to typical becnhmark input

Didou09: ok

PR #184, opened 2019-09-11
update in setup.py

### LINKED ISSUE #183: Mac OSX - OpenMP - reduction
After much consideration, I am opening an issue here even though it is **not** a **tofu** error but a problem of configuration between OSX, Cython an OpenMP, which leads to **wrong** results (tests failing) even if there is no issue in **tofu** and no warning/error displayed.

The first issue I found is that with ``homebrew``'s ``gcc-8``, OpenMP reductions are not being run in parallel but repeated. 
Some example can be found here:

https://stackoverflow.com/questions/54776301/cython-prange-is-repeating-not-parallelizing
https://github.com/bearnshaw/mult-cython
https://github.com/cython/cython/issues/2316

lasofivec:
Fixes #183 

- Now we don't specify the `gcc` compiler for Mac OSX (or any OS). It will use the system's default compiler. This fixes the above mentioned Issue as well as prevent the need for a patch (on ITER servers)

- Passed **black** on the ``setup.py`` file.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,6 +15,7 @@ import Cython as cth
 from Cython.Distutils import build_ext
 from Cython.Build import cythonize
 import numpy as np
+import _updateversion as up
 
 from distutils.command.clean import clean as Clean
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -31,31 +32,22 @@
...
 if platform.system() == "Windows":
     is_platform_windows = True
 
-if platform.system() == "Darwin":
-    # make sure you are using Homebrew's compiler
-    os.environ['CC'] = 'gcc-8'
-    os.environ['CXX'] = 'g++-8'
-else:
-    os.environ['CC'] = 'gcc'
-    os.environ['CXX'] = 'g++'
```
lasofivec: @Didou09 here we used to force which `gcc` was being used. By default `homebrew`'s gcc is `gcc-8`.
Now we let the system's `gcc`

Issue #187, opened 2019-09-11
Two init functions for CoilPF


lasofivec:
In `_core.py` there are two different initializers for `CoilPF` class:

https://github.com/ToFuProject/tofu/blob/807da74401e4e96f849127e7dae8afcd558f9dc0/tofu/geom/_core.py#L1503-L1518

lasofivec:
It seems to me like we can keep the second one. @Didou09 if you confirm I can solve this issue

Didou09:
Thanks for the info, I'll look tomorrow morning also, not sure immediately

PR #192, opened 2019-09-13
Issue187 duplicate init for coils pf

Didou09:
Fixes, in devel, Issue #187 

Extra verbose tf.geom.CoilsPF.__init__() removed
Less verbose one kept, with mobile=False enforced


benchmarks/calc_signal_benchmark.py
```diff
@@ -42,9 +42,9 @@
 # Defining defaults
 ###################
 
-_LRES = [-3,-3,0]
-_LLOS = [5,5,0]
-_LT = [3,3,0]
+_LRES = [-3, -3, 0]
+_LLOS = [5, 5, 0]
+_LT = [3, 3, 0]
```
lasofivec: Useless but ok

Issue #186, opened 2019-09-11
In _core.py: unused variable


lasofivec:
https://github.com/ToFuProject/tofu/blob/807da74401e4e96f849127e7dae8afcd558f9dc0/tofu/geom/_core.py#L423-L425

The variable here is declared but not used

lasofivec:
I can erase this variable (and thus also `Typesu` which is only needed for `lc` declaration.
@Didou09  can you confirm it? If so, I can do it

Didou09:
Not sure, I need to think about it, it might be that there is typo on the line below and that it should be saying something like :
``` assert np.all(np.any(Typesu, axis=0)) ```

In which case lc and Typesu are necessary,
Give me one day more

lasofivec:
Since `Typesu` is the same as `Types` but after `np.unique` seems to me that both statements are equivalent.
Meaning `assert np.all(np.any(Typesu, axis=0))` is equivalent to `assert np.all(np.any(Types, axis=0))`, and the second one should be more general.

PR #194, opened 2019-09-13
[Issue 186] the assert was on lc, not Types

Didou09:
Fixes, in devel, Issue #186 

The assert was on lc, not on Types


Didou09:
Review not necessary

PR #195, opened 2019-09-13
Build docs with sphinx-gallery

flothesof:
Hi all,

this is a first draft of sphinx-gallery generation of examples.
Let me know what you think of it.

Regards,
Florian


flothesof:
I managed to get it to work, despite the windows specific problems. Among the fixes are things like this:
```
-            lsnvert = np.asarray(lsnvert, dtype=np.int64)
+            lsnvert = np.asarray(lsnvert, dtype=np.long)
```
There are probably some more of these to be found: they cause problems on Windows!

tofu/geom/_core.py
```diff
@@ -3665,7 +3665,7 @@ def _prepare_inputs_kInOut(self, D=None, u=None, indStruct=None):
                 else:
                     num_tot_structs += len(ss.Lim)
 
-            lsnvert = np.asarray(lsnvert, dtype=np.int64)
+            lsnvert = np.asarray(lsnvert, dtype=np.long)
```
Didou09: Not sure how this affects linux users, waiting for feedback from Laura

tofu/geom/utils.py
```diff
@@ -35,8 +36,8 @@
 _lok = np.array([_lok, _lok+10])
 
 _here = os.path.abspath(os.path.dirname(__file__))
-_root = _here[:_here.rfind('/tofu')]
-_path_testcases = os.path.join(_root,'tofu/geom/inputs')
+_root = tofu.__path__[0]
+_path_testcases = os.path.join(_root, 'geom', 'inputs')
```
Didou09: good practice, thanks

tofu/geom/_core.py
lasofivec: I'm gonna test it in my machine, could you try it also in CentOS ?

lasofivec:
Made changes for pep8, merged with `devel`, and tried on my Mac and Ubuntu. Should we wait for @Didou09 to try on CentOS?

lasofivec:
Forgot to mention: the example is 🔥

PR #196, opened 2019-09-13
[documentation] update of documentation

lasofivec:
- updated config file for documentation: automatically taking the version of tofu
- merged with devel
- `black`-ed python files


lasofivec:
What about issue #200?

PR #199, opened 2019-09-24
[bugfix/portability] handling long types on windows

lasofivec:
It's working now on my computer (I'll merge this to your fork first if you want @flothesof)
For the orginial PR #197

### DIFF
--- a/tofu/data/_core.py
+++ b/tofu/data/_core.py
@@ -46,7 +46,28 @@ _INTERPT = 'zero'
 #############################################
 
 def _format_ind(ind=None, n=None):
-    """Type checking routine for index variable."""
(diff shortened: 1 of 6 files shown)


lasofivec:
I think @flothesof comment on PR #197 was spot on.

> A better way to check this might be to convert the array to a numpy format rightaway and then inspect the dtype instead of the type?

_Originally posted by @flothesof in https://github.com/ToFuProject/tofu/pull/197#issuecomment-534496005_

I said that `np.long` is not a valid numpy type, see:

https://docs.scipy.org/doc/numpy/user/basics.types.html?highlight=long

However in older `numpy` versions this is not true

https://docs.scipy.org/doc/numpy-1.15.4/user/basics.types.html?highlight=long

PR #207, opened 2019-10-03
Example signal computation with custom emissivity

flothesof:
As discussed with @Didou09 this is a "new" example (inspired by one of the benchmarks).


flothesof:
I've fiddled a bit more with the plots and realized that I have questions about the following items:

- what are the coordinate conventions for tofu ? (X Y Z in most cases, but how is the usual coordinate system defined?) should we make this appear somewhere in the documentation (in a page dedicated to conventions used, or alternatively in the introductory examples?)
- I was unable to get the example working using no time-dependency (setting t=None) gave me this error, while I assumed this would work 
```
  File "tofu\geom\_GG03.pyx", line 3057, in tofu.geom._GG03.LOS_calc_signal
    val_2d = func(pts, t=t, **fkwdargs)
ValueError: Buffer has wrong number of dimensions (expected 2, got 1)
```

lasofivec:
> I've fiddled a bit more with the plots and realized that I have questions about the following items:
> 
>     * what are the coordinate conventions for tofu ? (X Y Z in most cases, but how is the usual coordinate system defined?) should we make this appear somewhere in the documentation (in a page dedicated to conventions used, or alternatively in the introductory examples?)

Probably we should do both suggested solutions. For sure a section on "tofu's conventions" is missing in the doc.

>     * I was unable to get the example working using no time-dependency (setting t=None) gave me this error, while I assumed this would work
> 
> 
> ```
>   File "tofu\geom\_GG03.pyx", line 3057, in tofu.geom._GG03.LOS_calc_signal
>     val_2d = func(pts, t=t, **fkwdargs)
> ValueError: Buffer has wrong number of dimensions (expected 2, got 1)
> ```

As stated in the doc, the function ``emissivity`` should always return a 2D array. I fixed that directly in your example.

https://github.com/ToFuProject/tofu/blob/93ae5c51c4a114c869af53b531ae300bf310a6db/tofu/geom/_GG02.pyx#L2882-L2888

However the tutorial is still failing, this time when plotting the signal. I opened an issue regarding this, see ...

Didou09:
Problem solved, see Issue #217 
Tip though : the user needs to click once on the time axis

lasofivec:
This example is for the online documentation, so the figures will be non interactive.
Could you add a comment in the test with what you mean?

Didou09:
Plotting issue solved,
But bug found in _GG => opened issue #247 for @lasofivec , high priority

lasofivec:
I'm going to merge this pull request since all bugs have been corrected in branch `Issue247-bisect` and are independent of this branch :)

Issue #209, opened 2019-10-07
Plasma2D interpolation failing


Didou09:
Affect tofu 1.4.1:

Minimum working example:

```
multi = tf.imas2tofu(shot=54719, ids=['wall', 'pulse_schedule', 'soft_x_rays', 'equilibrium'])
sxr = multi.to_Data('sxr')
cam = sxr.lCam[0]
cam.set_dsino([2.4, 0.])
pts = cam.dsino['pts']
plasma = multi.to_Plasma2D()

plasma.interp_pts2profile(pts=pts, quant='equilibrium.rhotn')
```

Didou09:
After Jorge's investigations it seems the error may come from:
* Wrong assert (not...)
* tall not defined
* wrong default value (None instead of 'nearest')

To be verified

Didou09:
Fixed, in devel, by PR #210

PR #210, opened 2019-10-08
Issue209 plasma2 d interpolation

Didou09:
The bug was coming from interp_t not being set


Didou09:
Fixes, in devel, issue #209

lasofivec:
done

Issue #221, opened 2019-10-29
sdist and PyPi tarball is not clean.


Louwrensth:
The [PyPi package for 1.4.1](https://pypi.org/project/tofu/#files), contains the spurious build products, this is the diff after running ``python setup.py clean`` in the PyPi tarball:
```
[vandell@hpc-login03 ToFu]$ git diff --no-index --stat tofu-1.4.1-orig/ tofu-1.4.1/
 /dev/null => tofu-1.4.1/__pycache__/_updateversion.cpython-36.pyc |    Bin 0 -> 1064 bytes
 tofu-1.4.1-orig/tofu/geom/_GG02.cpp => /dev/null                  | 104230 ----------------------------------------------------------------------------------------------------------------------------------
 /dev/null => tofu-1.4.1/tofu/geom/_GG03.html                      |  64010 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 tofu-1.4.1-orig/tofu/geom/_basic_geom_tools.c => /dev/null        |  23457 ------------------------------
 tofu-1.4.1-orig/tofu/geom/_distance_tools.c => /dev/null          |  32225 ----------------------------------------
 tofu-1.4.1-orig/tofu/geom/_raytracing_tools.c => /dev/null        |  39715 --------------------------------------------------
 tofu-1.4.1-orig/tofu/geom/_sampling_tools.c => /dev/null          |  34027 -------------------------------------------
 tofu-1.4.1-orig/tofu/geom/_vignetting_tools.cpp => /dev/null      |  22916 -----------------------------
 8 files changed, 64010 insertions(+), 256570 deletions(-)
```

This causes problems when building ToFu with ``pip`` (as done by EasyBuild), because it will not rebuild these build products. You'll get errors like ``AttributeError: module 'tofu' has no attribute 'geom'``

See: 
#221 
#201 
https://jira.iter.org/browse/IMAS-2613
https://github.com/Louwrensth/easybuild-easyconfigs/commit/9adda488f7930fbf226dc8fcf1b0db2400c74037#commitcomment-35701902

I'll create a fix for this.

PR #222, opened 2019-10-29
Setup.py should not cythonize extensions except when building. #221

Louwrensth:
Cython’s build_ext module runs cythonize as part of the build process:
https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#distributing-cython-modules

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -414,7 +414,7 @@ setup(
     #    ],
     # },
     ext_modules=extensions,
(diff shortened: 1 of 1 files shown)


lasofivec:
I wasn't aware of this. Thank you !

Didou09:
Thanks @lasofivec for taking care of this, 
Are we sure this R passed all unit tests ? If not do we know why / what ?

Issue #224, opened 2019-10-30
Windows setup's path of README file


munechika-koyo:
When I install tofu into Windows PC using setup.py, it doesn't work well and I found the bugs.
So, I try to get rid of them in the `setup.py` file.

munechika-koyo:
It seems that they have something to do with getting the path and typo.

lasofivec:
Could you please copy paste the error ?
As well as the command that you are using to compile the library (eg. ``python setup.py install`` or ``python setup.py build_ext --inplace``) ?

munechika-koyo:
I tried both of them, and the same error message appeared:
```
$ python setup.py install
cython version = 0.29.12
numpy  version = 1.16.4
cython version = C:\Users\Koyo\Anaconda3\lib\site-packages\Cython\__init__.py
numpy  version = C:\Users\Koyo\Anaconda3\lib\site-packages\numpy\__init__.py
................ checking if openmp installed...
................ checking if openmp installed... >  False

Version for setup.py :  1.4.0-745-ge908e77

Traceback (most recent call last):
  File "setup.py", line 221, in <module>
    _README = [ff for ff in os.listdir(os.path.abspath(__file__))
NotADirectoryError: [WinError 267] ▒f▒B▒▒▒N▒g▒▒▒▒▒▒▒▒▒▒▒ł▒▒B: 'D:\\Documents\\tofu\\setup.py'
```
sorry that the text is garbled because that was caused by Japanese font.
But, it means that " the following path does not exist".

I think it is caused by using `os.listdir(os.path.abspath(__file__)`. This seems to make something wrong in windows.
I personally recommend using `os.get_cwd`

flothesof:
Hi there!
I'm surprised by @munechika-koyo 's error.
According to the docs, the recommended way of installing is pip install -e .[dev] 
Does that work better on your system? 
This works without a problem on my machine in fact. 
Regards 
Florian

Didou09:
On the one hand, @flothesof is right, for windows users, pip is the recommended way to go.
On the other hand, @munechika-koyo, as a contributor, allows us to test tofu installation in various configurations, and the 'python setup.py' should work, in principle...

We should try to make sure it works, installing tofu should be as easy as possible no matter how it is installed (conda, pip, python setup.py...)

However, @munechika-koyo , I cannot agree on your coding suggestion because both commands do not give the same result:

Here we want get the directory of the **file** (i.e. in which the module being executed in located)
[...]

PR #225, opened 2019-10-30
Correct path for README files search in setup

munechika-koyo:
I changed a little code to get rid of bug.
I would appreciate it if you could make sure it is fine and merge it to master branch. 

_Laura's comments_
This pull requests could be divided into two:
- Solving issue #224 : fixed path of `README` file in `setup.py`
- Introducing new ITER geometry "B4"


lasofivec:
Hi Koyo, thanks for the PR.

> I changed a little code to get rid of bag.

What do you mean with "bag"? 

A couple of remarks: 

- please always do the pull requests to `devel`. The `master` branch will be only changed around twice per year to maintain the stability of the library.
- If you see the comments above, you will see a automatic bot called `pep8-speaks` that comments on your coding style. I am aware that here it tells you a remark from code that you didn't actually code, and this is due to the fact that we didn't have this tool before and that there are still some areas where the coding style is still not perfect. But just as an exercise: would you mind correcting these errors. Here,  `pep8-speaks` found the following issues:

> In the file tofu/geom/utils.py:
> Line 629:25: E231 missing whitespace after ':'

if we see the line 629:

`            'A2': {'Exp':_ExpITER,`

it should be changed to:

`            'A2': {'Exp': _ExpITER,`

(notice the withespace after the `:`)

- I updated the documentation: see mostly the [contributing section](https://tofuproject.github.io/tofu/contributing.html). I precisely added some comments on the `devel`/`master` comments and on the `pep8-speaks`

tofu/geom/utils.py
```diff
@@ -622,10 +622,11 @@ def _compute_CamLOS2D_pinhole(P=None, F=0.1, D12=0.1, N12=100,
 ###########################################################
 
 _ExpWest = 'WEST'
+_ExpITER = 'ITER'
 
 _dconfig = {'A1': {'Exp':_ExpWest,
                    'Ves': ['V1']},
-            'A2': {'Exp':'ITER',
+            'A2': {'Exp':_ExpITER,
```
lasofivec: this is better ! thanks !

lasofivec:
I opened issue #227

setup.py
```diff
@@ -234,9 +234,10 @@ def get_version_tofu(path=_HERE):
 
 # Get the long description from the README file
 # Get the readme file whatever its extension (md vs rst)
+
 _README = [
     ff
-    for ff in os.listdir(os.path.abspath(os.path.dirname(__file__)))
```
Didou09: In principle, this should not change the result, since _HERE = os.path.abspath(os.path.dirname(__file__))
as defined above (line 220)

Does it really change something ? If so, does someone understand why ?

lasofivec: it doesn't but it's better

Didou09: 👍

lasofivec:
@Didou09 @munechika-koyo since now there is a branch specific to the new geometry addition `Issue220_NewITERConfig` I feel like this PR should only include the changes to the setup.py.
Event if it will be a small change, I think is good practice to separate the different changes on different PRs.
What do you think ? I can make the changes and close the PR if you argee

Didou09:
I totally agree, let's separate into elementary changes, it's way cleaner

munechika-koyo:
I raised the new PR https://github.com/ToFuProject/tofu/pull/245.
Is this fine?

lasofivec:
Yes thank you Koyo !!
I'll correct this branch and merge it.

PR #226, opened 2019-10-31
Br/add more field lines

jmoralesFusion:
It will not pass pep8 sorry I will modify next week


Didou09:
Thanks @jmoralesFusion for this PR,
before we validate, can you just take the time to address the PEP8 compliance (we are trying to get tofu gradually uniformized in terms of coding style and conventions). A bit cumbersome the first few times but then we'll all get the good practices :-)

Didou09:
The units tests are all passing for python 3.6, which is sufficient because we'll drop python 2.7 from now, so it's fine

tofu/utils.py
```diff
@@ -677,11 +677,33 @@ def load_from_imas(shot=None, run=None, user=None, tokamak=None, version=None,
...
+
+        multi = imas2tofu.MultiIDSLoader(shot=shot[0], run=run, user=user,
+                                        tokamak=tokamak, version=version,
+                                        ids='equilibrium')
+        equi = multi.to_Plasma2D()
+
         if t is None:
-            t = np.r_[38]
+            # Get time in the middle of equilibrium time interval
+            t = equi.ddata['equilibrium.t']['data'][int(0.5*equi.ddata['equilibrium.t']['data'].size)]
```
Didou09: good idea

tofu/mag/test_magFieldLines.py
```diff
@@ -0,0 +1,20 @@
...
+    TEST magFieldLines
+'''
+from __future__ import (unicode_literals, absolute_import,  \
+                        print_function, division)
+import doctest
+import os
+import sys
+
+#print('path 1 =', sys.path)
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
```
Didou09: Please make sure any change to the system path is exceptional and temporary
The system path should not be changed without the user knowing, unless it is temporary (i.e.: only in for a limited time inside the file)

=> please restore the original system path as soon as possible after the tests are run

=> Actually, this is not necessary, all unit tests should be placed in tofu/tests/

=> Temporary fix: restore original system path in the file, I validate the PR, then we open a new issue to move all unit tests to the proper repo (we will create a dedicated tofu/tests/tests09_mag/ repo, and get the template from other unit tests

tofu/mag/test_magFieldLines.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# vim: tabstop=4 expandtab shiftwidth=4 softtabstop=4
+# Also if needed: retab
```
lasofivec: seems like this is specific to your vim config, and not necessary to add ?

tofu/mag/test_magFieldLines.py
```diff
@@ -0,0 +1,20 @@
...
+import sys
+
+#print('path 1 =', sys.path)
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
+#print('path 2 =', sys.path)
+
+#import mag
+from mag import magFieldLines
+
+doctest.testmod(magFieldLines, verbose=True)
```
lasofivec: Not sure how @Didou09 sees this, but... I feel like this test should probably be moved to the `tests` directory?
Whatever the solution chosen, it should take into account that we are going to get rid of `nose`. 
See #232

tofuplot.py
```diff
@@ -143,6 +144,14 @@ def _str2bool(v):
                         nargs='+', default=None)
     parser.add_argument('-t0', '--t0', type=str, required=False,
                         help='Reference time event setting t = 0', default=_T0)
+    parser.add_argument('-t', '--t', type=float, required=False,
+                        help='Input time when needed')
```
lasofivec: @Didou09 seems like replacing renaming a flag that you used to be used for something else, and adding a new one  (type) for the old one (tokamak), can lead to confusion. Could the long flag for the time could be at least `--time`, or even better `--input-time`?

tofu/mag/test_magFieldLines.py
Didou09: Agreed, anything specific to a contributor's working environment shall be removed

tofu/mag/test_magFieldLines.py
Didou09: As commented in the review, I suggest:
* Step 1:  minimize changes to system path and restore original system path before the end of the tests, I think we can validate the PR from there
* Step 2: Move the unit tests to a dedicated repo in tofu/tests (create a dedicated issue and PR)
* Step 3: move from nose to something else (see issue #232, but we are not there yet)

tofuplot.py
Didou09: Discussed together, name flag -tok (short version) is actually probably better for --tokamak and -t for time is also more intuitive.
So it's fine for me.

The question on the long version (i.e.: --t vs --time) remains, I'm not sure what to think, because updating to --time would also require updating the corresponding keyword argument of the python function.

I suggest : leave it as is (--t) for now, and we'll see later if users mention any issue / ambiguity with that

lasofivec: fine by me :)

jmoralesFusion:
Hello pep8 now ok except for test files but they will be removed for a test folder when test package will be chosen. And for the _core.py file I didn't touch it so pep8speaks bug? 
Hope everything is fine for an alpha release.

Didou09:
Fixes in devel Issue #235

Issue #228, opened 2019-11-04
Change default separator in to_dict() from '_' to '.'


Didou09:
The default separator for flattening nested dict is currently `'_'`.

* It keeps users from defining attributes with `'_'` in the name
* A `'.'` would better render te nested structure

=> Switch from `'_'` to `'.'`
=> Ensure backward-retrocompatibility in `tf.load()` by implementing a general test

Didou09:
Fixed in devel by PR #229

PR #229, opened 2019-11-04
Issue228 default separator to dict

Didou09:
Fixes in devel issue #228 

Default separator is now '.' instead of '_'
=> renders better the nested dict structure
=> allows to use '_' in attributes names


Didou09:
Only python 2.7 not passing
=> merge because we are dropping python 2.7 (cf. PR #231 )

Didou09:
Agreed, I'll try to stick to that from now on

Issue #241, opened 2019-11-05
Bug in Rays._complete_dX12() on WEST / IMAS


Didou09:
MWE on IRFM / spica:

import tofu as tf
multi = tf.imas2tofu.MultiIDSLoader(shot=53923, ids=['pulse_schedule', 'wall', 'equilibrium', 'bolometer'])
bolo = multi.to_Data('bolometer')

Raises an error:

PR #242, opened 2019-11-05
Bug was not there, but __repr__() now more robust in debug

Didou09:
Erratum
---------
The bug in `Rays._complete_dX12()` (see Issue #241) was not a bug: it was the input that was inappropriate (the line of sight geometry was not available in IDS). 
=> No change to Rays class

More informative camera construction from IDS
-----------------------------------------------------
A warning has been added to camera construction from ids if some channels have nan in D or u

General Improvement
---------------
I noticed that the overloading of the parent class ToFuObjectBase.__repr__() method, though very useful when everything works fine, was a problem when in debug mode inside a class because the debugger wants to print the current variables, including **_self_**. But while debugging inside the class, **_self_** is not fully valid yet and the get_summary() medthod may not work, leading to the debugger bugging...

The parent method __repr__() is now more robust in debug mode: falls back to __class__.__name__ if the current instance is not valid.

Issues
-------

Fixes, in devel, Issue #241

### DIFF
--- a/tofu/imas2tofu/_core.py
+++ b/tofu/imas2tofu/_core.py
@@ -2699,7 +2699,14 @@ class MultiIDSLoader(object):
             u = np.array([oo[:,1,0]*np.cos(oo[:,1,2]),
                           oo[:,1,0]*np.sin(oo[:,1,2]), oo[:,1,1]])
             u = (u-D) / np.sqrt(np.sum((u-D)**2, axis=0))[None,:]
-            dgeom = (D,u)
+            dgeom = (D, u)
+            indnan = np.any(np.isnan(D), axis=0) | np.any(np.isnan(u), axis=0)
+            if np.any(indnan):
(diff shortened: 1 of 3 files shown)


lasofivec:
> ## Erratum
> 
> The bug in `Rays._complete_dX12()` (see Issue #241) was not a bug: it was the input that was inappropriate (the line of sight geometry was not available in IDS).
> => No change to Rays class

I'm confused as to why mention that in this PR ?
If the issue wasn't an issue, comment there and close it ?
And thus this PR doesn't solve the Issue as it wasn't an issue...

Didou09:
Agreed, 
issue #241 commented and closed accordingly
This PR will simply serve as reference for it, but you are right, next time good practice would be to comment in issue and close it, since the PR doesn't change anything on this particular point

Didou09:
agreed

Didou09:
done

PR #245, opened 2019-11-05
Issue220 new iter config

munechika-koyo:
New ITER configuration is added to tofu as a case 'B4' default one.


Didou09:
@lasofivec , are you correcting the last 3 PEP8 non-conformities or do we consider this is ok ?

lasofivec:
@Didou09, I would say yes :)
(sorry I think we were reviewing the doc at the same time)
lines too long 80 instead of 79, I think that's fine

Didou09:
Nice, congratulations Koyo for your first merged PR !
And thanks Laura for helping a lot !

PR #246, opened 2019-11-05
Benchmarks now with py3

lasofivec:
Changes in benchmarks scripts:

- instead of `print(..., file=...)` now using the more elegant `write(...`
- took out the `__init__.py` file so that the directory is not packages (only useful for devs)
- made default test smaller


Didou09:
> * took out the `__init__.py` file so that the directory is not packages (only useful for devs)
> * made default test smaller

2 good ideas :-)

lasofivec:
> Ok for me, seems cleaner
> Are you sure with the f.write(...) ? I agree it is more elegant and pythonic but it seems quite less compact.
> I'm ok with both, it's just because you raised the point and we might try to do the same in the future based on this discussion.
> 
> Did you try to launch the benchmark again after these changes ?
> Since they are not in the unit tests, travis does not check them (unless my memory is failing)

Yes I tried them, to be on the safe side, could you try them as well ?

Didou09:
Ok, I coul try them on the IRFM clusters, it works too 👍

Issue #252, opened 2019-11-15
Re-format output of user-defined signal function when t=None


lasofivec:
From #248 and https://github.com/ToFuProject/tofu/pull/207#issuecomment-540046272, I've gathered that the doc for how to define the function of the signal to integrate is not read so when "t=None", there is an error that the user does not understand.
I think it will be easier to check what's being passed when the pre-formatting is done, and use a wrapper so that the function always return a 2D array.

Didou09:
Fixed in devel by PR #254

Issue #248, opened 2019-11-08
assert ndim in [2,3] error in _checkformat_inputs_ddataRef


jdhare:
Hello,

I have generated our cameras for ITER, and I am trying to calculate the signal using calc_signal()

Even when using a dummy synthetic emission function such as: 
`def ff(Pts, t=None):
    return np.zeros(Pts.shape[1])+1
`
I still get the following error:

`c.calc_signal(ff, t=None)
Traceback (most recent call last):
  File "/home/jhare/.conda/envs/tofu/lib/python3.6/site-packages/IPython/core/interactiveshell.py", line 3326, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-110-851451bbde5c>", line 1, in <module>
    c.calc_signal(ff, t=None)
  File "/home/jhare/.conda/envs/tofu/lib/python3.6/site-packages/tofu/geom/_core.py", line 4227, in calc_signal
    osig = tfd.DataCam1D(**kwdargs)
  File "/home/jhare/.conda/envs/tofu/lib/python3.6/site-packages/tofu/data/_core.py", line 160, in __init__
    super(DataAbstract,self).__init__(**kwdargs)
  File "/home/jhare/.conda/envs/tofu/lib/python3.6/site-packages/tofu/utils.py", line 719, in __init__
    self._init(**kwdargs)
  File "/home/jhare/.conda/envs/tofu/lib/python3.6/site-packages/tofu/data/_core.py", line 610, in _init
    self._set_ddataRef(**kwddataRef)
  File "/home/jhare/.conda/envs/tofu/lib/python3.6/site-packages/tofu/data/_core.py", line 631, in _set_ddataRef
    lout = self._checkformat_inputs_ddataRef(**kwdargs)
  File "/home/jhare/.conda/envs/tofu/lib/python3.6/site-packages/tofu/data/_core.py", line 276, in _checkformat_inputs_ddataRef
    assert ndim in [2,3]
AssertionError`

The camera seems otherwise fine, I can plot its LOS on the vessel etc. Any ideas?

Didou09:
Yes, the emissivity function should always return a 2d array, where the first dimension is for time evolution.
Try redefining it as 
`def ff(Pts, t=None): return np.zeros((1, Pts.shape[1]))+1 `

It should solve your problem.

A caveat though: we very recently identified a bug in some the signal computation routines (see Issue #247 for more details).
[...]

PR #254, opened 2019-11-15
Feature #252: clearer error and check for emissivity function

lasofivec:
Cleaner errors when user didn't define a correct emissivity function.

For example, if using function in https://github.com/ToFuProject/tofu/issues/248#issue-520034648, we get:

> AssertionError: Input emissivity function (ff):  When t=None or t is a scalar, ff must return a 2D (1, npts) np.ndarray!

Solves #252


Didou09:
Good idea,
Just taking over from here to:
* finish PEP8 compliance
* Modify a bit the error handling (see review)

tofu/geom/_GG.pyx
```diff
@@ -2771,74 +2768,6 @@ def LOS_get_sample(int nlos, dL, double[:,::1] los_lims, str dmethod='abs',
 ######################################################################
 #               Signal calculation
 ######################################################################
-cdef get_insp(ff):
```
Didou09: Ok, so I understand you are taking the function checking out of _GG to place into _core right ?
That's fine for me, just want to be sure I get it ;-)

tofu/geom/_core.py
```diff
@@ -5760,6 +5760,86 @@ def calc_min_rho_from_Plasma2D(self, plasma, t=None, log='min',
...
+    def get_inspector(self, ff):
+        out = inspect.signature(ff)
+        pars = out.parameters.values()
+        na = np.sum([(pp.kind == pp.POSITIONAL_OR_KEYWORD
+                      and pp.default is pp.empty) for pp in pars])
+        kw = [pp.name for pp in pars if (pp.kind == pp.POSITIONAL_OR_KEYWORD
+                                         and pp.default is not pp.empty)]
+        return na, kw
+
+    def check_ff(self, ff, t=None, ani=None):
```
Didou09: All good to me, the more detailed the error messages, the better it is for everyone indeed  :-)

tofu/geom/_core.py
```diff
@@ -5760,6 +5760,86 @@ def calc_min_rho_from_Plasma2D(self, plasma, t=None, log='min',
...
+        time_steps = -1
+        # .. Checking basic definition of function ..........................
+        str_error = "Input emissivity function (ff): "
+        assert hasattr(ff, '__call__'), (str_error
```
Didou09: However, small caveat:  the sue of assert, though compact, is not the best option in my sense.
If statements would be bit better because:
* more readable code structure
* The error messages are defined only if the check is not passed (with assert, the message has to be defined even if there is no error).
* This [technical reason](https://softwareengineering.stackexchange.com/questions/225956/python-assert-vs-if-return)

I'll take over from here just to turn asserts into if, for good habits

tofu/geom/_core.py
```diff
@@ -5760,6 +5760,86 @@ def calc_min_rho_from_Plasma2D(self, plasma, t=None, log='min',
...
+        assert t is None or is_t_type_valid, (str_error
+                                              + "Arg t must be None,"
+                                              + " a scalar or an iterable!")
+        # .. Testing outputs ...............................................
+        test_pts = np.array([[1, 2], [3, 4], [5, 6]])
+        npts = test_pts.shape[1]
+        try:
+            out = ff(test_pts, t=t)
+        except Exception:
+            assert False, (str_error
```
Didou09: To force raise an exception just use
`raise Exception(msg)`

(assert False if overkill)

tofu/geom/_core.py
```diff
@@ -5760,6 +5760,86 @@ def calc_min_rho_from_Plasma2D(self, plasma, t=None, log='min',
...
+                           + " must take one positional arg:"
+                           + " a (3, npts) np.ndarray")
+        if hasattr(t, '__iter__'):
+            time_steps = len(t)
+            err_msg = (str_error
+                       + " ff(Pts, t=t), where Pts is a (3, npts) np.array and"
+                       + " t a len()=nt iterable,"
+                       + " must return a (nt, npts) np.ndarray!")
+            assert (type(out) is np.ndarray
```
Didou09: Good practice : try using `isinstance(var, class)` instead of `type(var) is class` whenever possible 
(I used to do the same, sorry)

tofu/geom/_core.py
```diff
@@ -5760,6 +5760,86 @@ def calc_min_rho_from_Plasma2D(self, plasma, t=None, log='min',
...
+                           + " and t is a list (nt,)")
+                assert (type(out) is np.ndarray
+                        and out.shape == (time_steps, npts)), err_msg
+            else:
+                msg = (str_error
+                       + "If ani=True, ff must return a (1, npts)"
+                       + " np.ndarray when Pts is (3, npts), vect is"
+                       + " provided and t is None or a scalar")
+                assert type(out) is np.ndarray and out.shape == (1, npts), msg
+        return
```
Didou09: This function should return is_ani

tofu/geom/_core.py
lasofivec: ok, the code was originally yours :)

tofu/geom/_core.py
Didou09: Because:
* It used to in the original version
* the default value of ani is None, but the functions called later take ani as input and expect it to be True or False. This function function (check_ff) determines whether ani is True or False, even when it is None. It checks and format ani for other functions called later

Issue #255, opened 2019-11-18
Possible bug / inconsistency in _GG.LOS_calc_signal()


Didou09:
While trying to finish / polish the PR #254 , and testing all tutorials, an error occured for:

tutorial plot_custom_emissivity.py
minimize = 'hybrid'
method = 'sum'

Upon closer inspection, it comes from line 3132 in _GG.py.
It seems method = 'sum' is the only case where the result is stored in sig_mv instead of sig.
I suspect this is the source of the error..
Indeed, there is no error if method  = 'simps' or method = 'romb'

The error-check matrix is

|      | sum | simps | romb |
|----|------|--------|-------|
hybrid | Error | ok | ok
calls | ok | ok | ok
memory | ok | ok | ok

The error message is:
```
Traceback (most recent call last):
  File "/Applications/Anaconda/python36/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Home/DV226270/ToFu_All/tofu_git/tofu/tofu/tests/tests09_tutorials/tests01_runall.py", line 182, in test03_plot_custom_emissivity
    self._test_tuto('tuto_plot_custom_emissivity')
  File "/Home/DV226270/ToFu_All/tofu_git/tofu/tofu/tests/tests09_tutorials/tests01_runall.py", line 164, in _test_tuto
    raise Exception(msg)
Exception: failed to get the current screen resources
Traceback (most recent call last):
  File "/Home/DV226270/ToFu_All/tofu_git/tofu/tuto_plot_custom_emissivity.py", line 92, in <module>
    t=time_vector)
  File "/Home/DV226270/ToFu_All/tofu_git/tofu/tofu/geom/_core.py", line 6066, in calc_signal
    Test=True,
  File "tofu/geom/_GG.pyx", line 3132, in tofu.geom._GG.LOS_calc_signal
TypeError: only size-1 arrays can be converted to Python scalars
```

@lasofivec can you have a look on a dedicated branch ?

Didou09:
Important update:

When running the tutorial, I realize the error seems to be linked to the way the emissivity function is defined:

instead of using emissivity as input function we use:
`ff = lambda pts, t=None: np.zeros((1, pts.shape[1])) +1.0 if t is None else np.zeros((len(t), pts.shape[1])) +1.0
`
There is no error showing up...

What is strange is that both function are defined correctly, their output have the same shape:

```
[...]

PR #256, opened 2019-11-18
Feature #252 bisect

lasofivec:
Added a wrapper to the user defined function in the case t=None (or $t \in R$), such that there is no error raised and the output is reshaped.


lasofivec:
Tests are failing because of Issue #255 solved in another PR

PR #257, opened 2019-11-18
[bug fix #255] found bug

### LINKED ISSUE #255: Possible bug / inconsistency in _GG.LOS_calc_signal()

lasofivec:
Fixes #255

### DIFF
--- a/tofu/geom/_GG.pyx
+++ b/tofu/geom/_GG.pyx
@@ -2860,7 +2860,6 @@ def LOS_calc_signal(func, double[:,::1] ray_orig, double[:,::1] ray_vdir, res,
     cdef double[1] loc_eff_res
     cdef double[::1] reseff_mv
     cdef double[::1] res_mv
-    cdef double[::1,:] sig_mv
     cdef double[:,::1] val_mv
     cdef double[:,::1] pts_mv
     cdef double[:,::1] usbis_mv
@@ -2928,7 +2927,6 @@ def LOS_calc_signal(func, double[:,::1] ray_orig, double[:,::1] ray_vdir, res,
     n_imode = _st.get_nb_imode(imode)
     # Initialization result
     sig = np.empty((nt, nlos), dtype=float, order='F')
-    sig_mv = sig
     # If the resolution is the same for every LOS, we create a tab
     if res_is_list :
         res_arr = np.asarray(res)
@@ -3025,11 +3023,12 @@ def LOS_calc_signal(func, double[:,::1] ray_orig, double[:,::1] ray_vdir, res,
                     pts, usbis = _st.call_get_sample_single_ani(lims[0, ii],
                                                                 lims[1, ii],
                                                                 res_mv[ii],
-                                                                n_dmode, n_imode,
+                                                                n_dmode,
+                                                                n_imode,
                                                                 &loc_eff_res[0],
                                                                 &nb_rows[0],
-                                                                ray_orig[:,ii:ii+1],
-                                                                ray_vdir[:,ii:ii+1])
+                                                                ray_orig[:, ii:ii+1],
+                                                                ray_vdir[:, ii:ii+1])
                     # loop over time for calling and integrating
                     for jj in range(nt):
                         val = func(pts, t=ltime[jj], vect=-usbis, **fkwdargs)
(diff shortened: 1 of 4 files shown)


tofu/geom/_core.py
```diff
@@ -5815,7 +5817,7 @@ def check_ff(self, ff, t=None, ani=None):
         nt = len(t) if hasattr(t, '__iter__') else 1
 
         # .. Test anisotropic case .......................................
-        if ani is None:
+        if ani is False:
```
Didou09: I don't get it, if ani is False it means the user wants to force it to False, we should not change it.
We should only change it if None (means the user did not specify it).

tofu/geom/_core.py
```diff
@@ -5969,7 +5971,7 @@ def calc_signal(
         self,
         func,
         t=None,
-        ani=None,
```
Didou09: Same comment as above, None means the user did not specify it, False means he wants to force it to False

lasofivec: updated

lasofivec:
> Good, but:
> 
> * Could you find the reason why an error was thrown with vect=None ? It should not

Please see the lines where I tagged you

tofu/geom/_GG.pyx
```diff
@@ -3129,7 +3128,7 @@ def LOS_calc_signal(func, double[:,::1] ray_orig, double[:,::1] ray_vdir, res,
                                                                 ray_orig[:, ii:ii+1],
                                                                 ray_vdir[:, ii:ii+1])
                     val_2d = func(pts, t=t, vect=-usbis, **fkwdargs)
-                    sig_mv[:, ii] = np.sum(val_2d, axis=-1)*loc_eff_res[0]
+                    sig[:, ii] = np.sum(val_2d, axis=-1)*loc_eff_res[0]
```
lasofivec: @Didou09 :
> Could you find the reason why an error was thrown with vect=None ? It should not

It doesn't anymore. Here is the only real change in this PR, that fixes the bug.

lasofivec: PS: if you see line 3319, it only happened if `ani=True, minimize=hybrid, method=sum`, and if you see line 2863 I took out the variable which was no longer need and verified all loops to see that this error was nowhere else.

tofu/tests/tests09_tutorials/tests01_runall.py
```diff
@@ -154,8 +154,10 @@ def _test_tuto(cls, tuto=None, pathtuto=_HERE, root=_TFROOT):
             out = subprocess.run(cmd, shell=True, check=False,
                                  stdout=subprocess.PIPE,
                                  stderr=subprocess.PIPE)
-            if 'error' in out.stderr.decode().lower():
-                error = out.stderr.decode()
+            ls = out.stderr.decode().split('\n')
+            if any(['Error' in ss and not 'ModuleNotFoundError' in ss
+                    for ss in ls]):
+                error = '\n'.join(ls)
         except Exception as err:
```
lasofivec: I was correcting this... thanks :)

PR #258, opened 2019-11-19
[py3] took out all tests of python version

lasofivec:
Took out all lingering `if sys.version[0]=='2':`

### DIFF
--- a/tofu/geom/_comp.py
+++ b/tofu/geom/_comp.py
@@ -11,11 +11,6 @@ import numpy as np
 import scipy.interpolate as scpinterp
 import scipy.integrate as scpintg
 
-if sys.version[0] == "3":
-    from inspect import signature as insp
-elif sys.version[0] == "2":
-    from inspect import getargspec as insp
-
 # ToFu-specific
 try:
     import tofu.geom._def as _def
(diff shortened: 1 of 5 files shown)


tofu/geom/_comp.py
```diff
@@ -11,11 +11,6 @@
 import scipy.interpolate as scpinterp
 import scipy.integrate as scpintg
 
-if sys.version[0] == "3":
-    from inspect import signature as insp
```
Didou09: If this is suppressed too, where is insp defined ?

tofu/geom/_comp.py
```diff
@@ -1027,18 +1022,15 @@ def LOS_calc_signal(
         D, u, dL, DL=DL, dLMode=dLMode, method=method, Test=Test
     )
     out = insp(ff)
```
Didou09: insp is called here

lasofivec: sorry ! you are right ! fixed

tofu/geom/_comp.py
```diff
@@ -10,11 +10,7 @@
 import numpy as np
 import scipy.interpolate as scpinterp
 import scipy.integrate as scpintg
-
-if sys.version[0] == "3":
-    from inspect import signature as insp
-elif sys.version[0] == "2":
-    from inspect import getargspec as insp
+from inspect import signature as insp
```
lasofivec: corrected !

Issue #292, opened 2019-11-27
Unable to install from source on Windows


flothesof:
Hi,
I'm trying to install from source with latest devel.
However, it doesn't work.
```
(tofu_dev) C:\Users\FL232714\Documents\GitHub\tofu>pip install -e ".[dev]"
Obtaining file:///C:/Users/FL232714/Documents/GitHub/tofu
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  ERROR: Command errored out with exit status 1:
   command: 'C:\Anaconda3\envs\tofu_dev\python.exe' 'C:\Anaconda3\envs\tofu_dev\lib\site-packages\pip\_vendor\pep517\_in_process.py' get_requires_for_build_wheel 'C:\Users\FL232714\AppData\Local\Temp\tmprlzaqcsh'
       cwd: C:\Users\FL232714\Documents\GitHub\tofu
  Complete output (16 lines):
  Traceback (most recent call last):
    File "C:\Anaconda3\envs\tofu_dev\lib\site-packages\pip\_vendor\pep517\_in_process.py", line 207, in <module>
      main()
    File "C:\Anaconda3\envs\tofu_dev\lib\site-packages\pip\_vendor\pep517\_in_process.py", line 197, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "C:\Anaconda3\envs\tofu_dev\lib\site-packages\pip\_vendor\pep517\_in_process.py", line 54, in get_requires_for_build_wheel
      return hook(config_settings)
    File "C:\Users\FL232714\AppData\Local\Temp\pip-build-env-dzkqyuzc\overlay\Lib\site-packages\setuptools\build_meta.py", line 146, in get_requires_for_build_wheel
      return self._get_build_requires(config_settings, requirements=['wheel'])
    File "C:\Users\FL232714\AppData\Local\Temp\pip-build-env-dzkqyuzc\overlay\Lib\site-packages\setuptools\build_meta.py", line 127, in _get_build_requires
      self.run_setup()
    File "C:\Users\FL232714\AppData\Local\Temp\pip-build-env-dzkqyuzc\overlay\Lib\site-packages\setuptools\build_meta.py", line 142, in run_setup
      exec(compile(code, __file__, 'exec'), locals())
    File "setup.py", line 21, in <module>
      import _updateversion as up
  ModuleNotFoundError: No module named '_updateversion'
  ----------------------------------------
[...]

PR #294, opened 2019-11-27
fix setup.py on Windows

flothesof:
This fixes the install problem I found in #292 when installing from source on Windows. 

As a side note, it seems the setup.py is starting to get bloated. In particular, I noticed that `get_version_tofu` and `updateversion` do something very similar (like checking if git is working in the path to know if we are on master...). This would be a an area for refactoring.


setup.py
```diff
@@ -142,9 +140,30 @@ def check_for_openmp(cc_var):
 
 
 # == Getting tofu version =====================================================
+
 _HERE = os.path.abspath(os.path.dirname(__file__))
-def get_version_tofu(path=_HERE):
 
+
+def updateversion(path=_HERE):
```
Didou09: Yes and no...
I agree having updateversion defined in a standalone side module is a bit cumbersome because it seems it is only needed in the setup.py.

But this was done in order to allow travis (or us) run updateversion as a standalone without having to import setup.py (which would run it).

updateversion can be useful as a standalone when we want to force an update of the version number prior to running the setup.py, using for example:

> python -c "import _updateversion as up; out = up.updateversion(); print(out)"

This is not possible anymore with this change.

Are you sure there is no other fix for issue #292 ?  It used to work before, I don't get why it stopped working

flothesof: Okay, thanks for reviewing. So it seems this should not be merged as such.
I believe then this needs more investigation. 
I have no idea why it's broken under Windows. Too bad...

Didou09: Nice,
Please also make sure the PR does not include the chages from PR #292 (i.e.: updateversion)
These are two independent changes, and as we discussed, I'm sure sure we can merge PR #292 as of now.
I suspect it is because you created your branch from the branch you had already created for PR #292 , right ?

Best practice : always branch from devel, a world where things are stable and validated, to make small incremental changes ;-)

lasofivec:
So I'm confused, is this solution working now ?

flothesof:
No, this is broken and will not be fixed with this PR. I think @Didou09 was talking about #295 which incorporated these changes. I rewrote my history on #295 to take them out. 

Closing since this will fix the problem.

PR #296, opened 2019-11-27
Fix issue295

### LINKED ISSUE #295: test failing on Windows
After installing from source, I'm running nosetests. 

Everything runs, except for one unit test.

```
  File "C:\Users\FL232714\Documents\GitHub\tofu\tofu\tests\tests09_tutorials\tests01_runall.py", line 20, in <module>
    _TFROOT = _HERE[:-_HERE[::-1].index('/tofu'[::-1])-len('/tofu')]
ValueError: substring not found
```

Will try to fix.

flothesof:
Fixes #295.

### DIFF
--- a/tofu/tests/tests09_tutorials/tests01_runall.py
+++ b/tofu/tests/tests09_tutorials/tests01_runall.py
@@ -17,7 +17,7 @@ from tofu import __version__
 
 
 _HERE = os.path.abspath(os.path.dirname(__file__))
-_TFROOT = _HERE[:-_HERE[::-1].index('/tofu'[::-1])-len('/tofu')]
+_TFROOT = os.path.abspath(os.path.join(os.path.dirname(_HERE), os.pardir))
 # _TFROOT = tf.__path__[0][:-5]
 # _PATHTUTO = os.path.join(_TFROOT, 'examples', 'tutorials')
 keyVers = 'Vers'
--- a/tofu/tests/tests09_tutorials/tests01_runall.py
+++ b/tofu/tests/tests09_tutorials/tests01_runall.py
@@ -1,25 +1,13 @@
-
-# External modules
 import os
 import shutil
-import types
 import subprocess
-import numpy as np
-import warnings as warn
 import matplotlib.pyplot as plt
-
-# Nose-specific
-from nose import with_setup     # optional
-
-# Importing package tofu.geom
-import tofu as tf
 from tofu import __version__
-
+from pathlib import Path
 
 _HERE = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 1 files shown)


tofu/tests/tests09_tutorials/tests01_runall.py
```diff
@@ -17,7 +17,7 @@
 
 
 _HERE = os.path.abspath(os.path.dirname(__file__))
-_TFROOT = _HERE[:-_HERE[::-1].index('/tofu'[::-1])-len('/tofu')]
+_TFROOT = os.path.abspath(os.path.join(os.path.dirname(_HERE), os.pardir))
```
Didou09: Smart move cowboy ! ;-)
But... although the use of os.pardir is a good idea, the result is not equivalent to the original.

Indeed, for example on my platform (where I have several nested /tofu directories):
```
In [7]: _HERE
Out[7]: '/Home/DV226270/ToFu_All/tofu_git/tofu/tofu/tests/tests09_tutorials'

In [8]: _TFROOT = _HERE[:-_HERE[::-1].index('/tofu'[::-1])-len('/tofu')]

In [9]: _TFROOT2 = os.path.abspath(os.path.join(os.path.dirname(_HERE), os.pardir))

In [10]: _TFROOT
Out[10]: '/Home/DV226270/ToFu_All/tofu_git/tofu'

In [11]: _TFROOT2
Out[11]: '/Home/DV226270/ToFu_All/tofu_git/tofu/tofu'
```

Here by _TFROOT, we mean the directory from which you should be able to import the local tofu (i.e. from which tofu is seen as a package), which one step upwards from your solution (in your case you're already inside the package itself, so you can't import it).

flothesof: Okay, I think I can update this one to get the desired behaviour.

tofu/tests/tests09_tutorials/tests01_runall.py
```diff
@@ -1,25 +1,13 @@
...
 import matplotlib.pyplot as plt
-
-# Nose-specific
-from nose import with_setup     # optional
-
-# Importing package tofu.geom
-import tofu as tf
 from tofu import __version__
-
+from pathlib import Path
```
Didou09: Nice, 
This however introduces a new dependency in tofu, except if pathlib is by default in Python 3.6 at least and supported in 3.7 too, is it ?

flothesof: It is, according to its doc, which mentions _New in version 3.4._

flothesof: https://docs.python.org/3/library/pathlib.html

flothesof:
I cleaned the history of this PR to just fix the unit test.

PR #297, opened 2019-11-27
New tutorial on builtin configs

flothesof:
I created a very short tutorial covering loading ITER, WEST and JET configs, as discussed today.

PR #298, opened 2019-11-27
Preparing tofu 1.4.2-a4

Didou09:
Main changes since last alpha:
---------------------------------
* New tutorial for new configs #297 
* pipy deployment in travis is 'sdist' (skipped if already exists and **then** bdist for MacOS only. #293 
* November meeting notes #291 
* Cleaned-up useless files #290 
* IMAS version is checked against latest version #289


Didou09:
I suggest we ignore PEP8 complains for tutorials, what do you think ?

PR #299, opened 2019-11-28
[Issue292] Hotfix: both solutions

Didou09:
Hotfix by defining updateversion in setup.py and keeping _updateversion.py also

Fixes, in devel, Issue #292

### DIFF
--- a/_updateversion.py
+++ b/_updateversion.py
@@ -1,16 +1,15 @@
 #!/usr/bin/env/python
 # coding=utf-8
 
-import sys
 import os
 import subprocess
 
-_HERE = os.path.join(os.path.abspath(os.path.dirname(__file__)),'tofu')
+_HERE = os.path.abspath(os.path.dirname(__file__))
 
 def updateversion(path=_HERE):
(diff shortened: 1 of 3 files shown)


Didou09:
Does seem to solve the problem, I could build a sdist locally and install it with pip on a fresh environment :-)

_updateversion.py
```diff
@@ -1,16 +1,15 @@
...
 import subprocess
 
-_HERE = os.path.join(os.path.abspath(os.path.dirname(__file__)),'tofu')
+_HERE = os.path.abspath(os.path.dirname(__file__))
 
 def updateversion(path=_HERE):
     # Fetch version from git tags, and write to version.py
     # Also, when git is not available (PyPi package), use stored version.py
-    version_py = os.path.join(path,"version.py")
+    version_py = os.path.join(path, 'tofu', 'version.py')
```
lasofivec: this change is independent form the issue right ?

Didou09: Yes, just making it cleaner :-)

Issue #307, opened 2019-11-29
Parallelization not working on ITER


Didou09:
On ITER clusters, using @munechika-koyo 's camera (which is big with 250,000 LOS) and an edge emissivity field from JINTRAC, the computation time (for res= 10 cm, just for testing) does not seem to be accelerated by parallelization:

Minimum working example (tofu 1.4.2-a5):
-----------------------------------------------
```
In [1]: import tofu as tf
/home/ITER/vezined/ToFu_All/tofu/tofu/__init__.py:95: UserWarning: 
The following subpackages are not available:
    - tofu.mag
  => see tofu.dsub[<subpackage>] for details.
  warnings.warn(msg)

In [2]: cam = tf.load('/home/ITER/munechk/public/MyTofu/output/ITER_test_camera_config.npz')
Loaded from:
    /home/ITER/munechk/public/MyTofu/output/ITER_test_camera_config.npz

In [3]: multi = tf.imas2tofu.MultiIDSLoader(user='hoeneno', tokamak='convert', shot=134000, run=29, ids=['core_sources', 'equilibrium', 'edge_sources'])
Getting ids   [occ]  tokamak  user     version  shot    run  refshot  refrun
------------  -----  -------  -------  -------  ------  ---  -------  ------
core_sources  [0]    convert  hoeneno  3        134000  29   -1       -1    
edge_sources  [0]    "        "        "        "       "    "        "     
equilibrium   [0]    "        "        "        "       "    "        "     

In [4]: _dshort = {'core_sources': {'1drhotn':'source[identifier.name=radiation].profiles_1d[time].grid.rho_tor_norm',
   ...:    ...:                             '1deEnergy':'source[identifier.name=radiation].profiles_1d[time].electrons.energy'
   ...:    ...:                             },
   ...:    ...:            'equilibrium': {'2dpsi': 'time_slice[time].profiles_2d[0].psi',
   ...:    ...:                            '2dmeshR': 'time_slice[time].profiles_2d[0].r',
   ...:    ...:                            '2dmeshZ': 'time_slice[time].profiles_2d[0].z'}}
   ...:                            

In [5]: multi.set_shortcuts(dshort=_dshort)

In [6]: plasma = multi.to_Plasma2D(shapeRZ=('R', 'Z'))
/home/ITER/vezined/ToFu_All/tofu/tofu/imas2tofu/_core.py:1972: UserWarning: The following data could not be retrieved:
	- equilibrium:
		2dB  : '2dBT'
[...]

PR #308, opened 2019-12-02
Faster los calc signal sum calls for newcalc=False

Didou09:
Motivations:
--------------
In the case where newcalc is False, we use a for loop over the LOS to compute the sum (integral).
I wanted to know whether there might be more efficient function available in numpy.
I asked the question on Stackoverflow.

It exists, and here it is:   np.add.reduceat()

Ref:
https://stackoverflow.com/questions/59079141/perform-numpy-sum-or-scipy-integrate-simps-on-large-splitted-array-efficient/59085492#59085492

This PR is interesting because the algorithm used when newcalc=False is very similar to the one used when (newcalc=True, method='sum', minimize='calls').
So acceleration in one case may be useful in the other.

Main changes:
----------------
* Replaced the for loop in Rays.calc_signal(newcalc=False) and Rays.calc_signal_from_Plasma2D(newcalc=False) by np.add.reduceat()
* While testing, a bug appeared in Plasma2D._get_indtu(), when t has only one time step. Fixed.

Benefits:
---------
A small benchmark on ITER (git + python setup.py build_ext) yielded (on a case with 250,000 LOS but only res=0.1):

```
In [1]: import tofu as tf
m/home/ITER/vezined/ToFu_All/tofu/tofu/__init__.py:95: UserWarning: 
The following subpackages are not available:
    - tofu.mag
  => see tofu.dsub[<subpackage>] for details.
  warnings.warn(msg)

In [2]: multi = tf.imas2tofu.MultiIDSLoader(user='hoeneno', tokamak='convert', shot=134000, run=29, ids=['edge_sources'])
Getting ids   [occ]  tokamak  user     version  shot    run  refshot  refrun
------------  -----  -------  -------  -------  ------  ---  -------  ------
edge_sources  [0]    convert  hoeneno  3        134000  29   -1       -1    

In [3]: _dshort = {'core_sources': {'1drhotn':'source[identifier.name=radiation].profiles_1d[time].grid.rho_tor_norm',
   ...:    ...:                             '1deEnergy':'source[identifier.name=radiation].profiles_1d[time].electrons.energy'
   ...:    ...:                             },
   ...:    ...:            'equilibrium': {'2dpsi': 'time_slice[time].profiles_2d[0].psi',
   ...:    ...:                            '2dmeshR': 'time_slice[time].profiles_2d[0].r',
   ...:    ...:                            '2dmeshZ': 'time_slice[time].profiles_2d[0].z'}}
   ...:                            

In [4]: multi.set_shortcuts(dshort=_dshort)

In [5]: plasma = multi.to_Plasma2D()

In [6]: cam = tf.load('/home/ITER/munechk/public/MyTofu/output/ITER_test_camera_config.npz')
sLoaded from:
    /home/ITER/munechk/public/MyTofu/output/ITER_test_camera_config.npz

In [7]: sig_oldbasic, units = cam.calc_signal_from_Plasma2D(plasma, quant='edge_sources.2dradiation', plot=False, res=0.1, newcalc=False, ufunc=False)

In [8]: sig_oldufunc, units = cam.calc_signal_from_Plasma2D(plasma, quant='edge_sources.2dradiation', plot=False, res=0.1, newcalc=False, ufunc=True, fill_value=0.)

In [9]: sig_newbasic, units = cam.calc_signal_from_Plasma2D(plasma, quant='edge_sources.2dradiation', plot=False, res=0.1, newcalc=True, method='sum', minimize='calls', ufunc=False)

In [10]: np.allclose(sig_oldbasic.data, sig_oldufunc.data)
Out[10]: True

In [11]: np.allclose(sig_newbasic.data, sig_oldufunc.data)
Out[11]: True

In [12]: %timeit sig_oldbasic, units = cam.calc_signal_from_Plasma2D(plasma, quant='edge_sources.2dradiation', plot=False, res=0.1, newcalc=False, ufunc=False)
11.8 s ± 90.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [13]: %timeit sig_oldufunc, units = cam.calc_signal_from_Plasma2D(plasma, quant='edge_sources.2dradiation', plot=False, res=0.1, newcalc=False, ufunc=True, fill_value=0.)
8.77 s ± 19.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [14]: %timeit sig_newbasic, units = cam.calc_signal_from_Plasma2D(plasma, quant='edge_sources.2dradiation', plot=False, res=0.1, newcalc=True, method='sum', minimize='calls', ufunc=False)
9.94 s ± 19.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Notice the 25% speed gain when newcalc=False.

The comparison with newcalc=True and a look at the code shows that in _GG.pyx, in the case (method='sum', minimize='calls'), the summing operation (l. 2991-2997) is done is a loop which is not parallelized, we could either:
* Parallelize this loop ?
* Quick hotfix : use the same function np.add.reduceat() to suppress the loop ?

What do you think @lasofivec ?

P.S.: keep in mind that currently, running tofu on the ITER clusters in ipython does not make use of the parallelization (cf. issue #307 ), which explains why in the above benchmark newcalc=True is not faster than newcalc=True, since it runs on only one CPU

### DIFF
--- a/tofu/data/_core.py
+++ b/tofu/data/_core.py
@@ -3677,11 +3677,15 @@ class Plasma2D(utils.ToFuObject):
         # Get indt (t with respect to tbinall)
         indt, indtu = None, None
         if t is not None:
-            indt = np.digitize(t, tbinall)
-            indtu = np.unique(indt)
+            if len(t) == len(tall) and np.allclose(t, tall):
+                indt = np.arange(0, tall.size)
+                indtu = indt
+            else:
+                indt = np.digitize(t, tbinall)
+                indtu = np.unique(indt)
+                # Update
+                tall = tall[indtu]
 
-            # Update
-            tall = tall[indtu]
             if idref1d is not None:
                 assert indtr1 is not None
                 indtr1 = indtr1[indtu]
--- a/tofu/geom/_core.py
+++ b/tofu/geom/_core.py
@@ -6025,7 +6025,7 @@ class Rays(utils.ToFuObject):
         reflections=True,
         coefs=None,
         ind=None,
-        out=object,
+        returnas=object,
         plot=True,
         dataname=None,
         fs=None,
@@ -6086,7 +6086,7 @@ class Rays(utils.ToFuObject):
         # Format input
 
         indok, Ds, us, DL, E = self._calc_signal_preformat(
-            ind=ind, DL=DL, out=out, Brightness=Brightness
+            ind=ind, DL=DL, out=returnas, Brightness=Brightness
         )
 
         if Ds is None:
@@ -6180,19 +6180,8 @@ class Rays(utils.ToFuObject):
             #    and interferometer)
             val = func(pts, t=t, vect=vect)
             # Integrate
-            if val.ndim == 2:
-                sig = np.full((val.shape[0], self.nRays), np.nan)
-            else:
-                sig = np.full((1, self.nRays), np.nan)
-            indpts = np.r_[0, indpts, pts.shape[1]]
-            for ii in range(0, self.nRays):
-                sig[:, ii] = (
-                    np.nansum(
-                        val[:, indpts[ii]:indpts[ii + 1]],
-                        axis=-1
-                    )
-                    * reseff[ii]
-                )
+            sig = np.add.reduceat(val, np.r_[0, indpts], axis=-1)*reseff[None, :]
+
         # Format output
         return self._calc_signal_postformat(
             sig,
@@ -6202,7 +6191,7 @@ class Rays(utils.ToFuObject):
             E=E,
             units=units,
             plot=plot,
-            out=out,
+            out=returnas,
             fs=fs,
             dmargin=dmargin,
             wintit=wintit,
@@ -6236,7 +6225,7 @@ class Rays(utils.ToFuObject):
         reflections=True,
         coefs=None,
         ind=None,
-        out=object,
+        returnas=object,
         plot=True,
         dataname=None,
         fs=None,
@@ -6250,7 +6239,7 @@ class Rays(utils.ToFuObject):
 
         # Format input
         indok, Ds, us, DL, E = self._calc_signal_preformat(
-            ind=ind, out=out, t=t, Brightness=Brightness
+            ind=ind, out=returnas, t=t, Brightness=Brightness
         )
 
         if Ds is None:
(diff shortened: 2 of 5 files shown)


tofu/data/_core.py
```diff
@@ -3677,11 +3677,15 @@ def _get_indtu(t=None, tall=None, tbinall=None,
         # Get indt (t with respect to tbinall)
         indt, indtu = None, None
         if t is not None:
-            indt = np.digitize(t, tbinall)
-            indtu = np.unique(indt)
+            if len(t) == len(tall) and np.allclose(t, tall):
```
Didou09: handle trivial case first, particularly suited for cases with t.size = 1

tofu/geom/_core.py
```diff
@@ -6025,7 +6025,7 @@ def calc_signal(
         reflections=True,
         coefs=None,
         ind=None,
-        out=object,
+        returnas=object,
```
Didou09: much more explicit variable name, especially since we tend to use 'out' to store temporary output inside functions.
It as the case in the other method by the way (calc_signal_from_Plasma2D()), out was redefined at line 6254

tofu/geom/_core.py
```diff
@@ -6180,19 +6180,9 @@ def calc_signal(
...
-            indpts = np.r_[0, indpts, pts.shape[1]]
-            for ii in range(0, self.nRays):
-                sig[:, ii] = (
-                    np.nansum(
-                        val[:, indpts[ii]:indpts[ii + 1]],
-                        axis=-1
-                    )
-                    * reseff[ii]
-                )
+            sig = np.add.reduceat(val, np.r_[0, indpts],
```
Didou09: More concise and numerically faster, only advantages :-)
Usable in _GG.LOS_calc_signal(method='sum', minimize='calls') ?
Usable elsewhere ?

Compatible with parallelization ?

tofu/geom/_core.py
```diff
@@ -6368,6 +6358,8 @@ def funcbis(*args, **kwdargs):
                     indpts[0], np.diff(indpts), pts.shape[1] - indpts[-1]
                 ]
                 vect = np.repeat(self.u, nbrep, axis=1)
+            if fill_value is None:
```
Didou09: Avoid nans, they are not handled by np.add.reduceat() => we set fill_value to 0 instead in this case (if not forced by the user)

tofu/geom/_core.py
lasofivec: nope, only functions that don't require gil can be parallelized, calling numpy functions requires the gil, so it is not possible. This is why it wasn't parallelized with newcalc=True

lasofivec:
It is not possible to parallelize the code that uses numpy function (GIL-requirement).
So I went for the "hotfix" which is not such a hotfix, more an enhancement.
I tried the new version, which seems to be working, I will create a new PR.

lasofivec:
Oh, one comment: in your benchmark, you put the parameter `ufunc = [True/False]` which is not a valid parameter for the function. I wonder if you were using the wrong version of tofu ? or a newer that was never pushed ?

Didou09:
It was a temporary version where all options were possible.
Since the becnhmark showed that using the ufunc (Universal Function, in this case np.add.reduceat) was significantly faster, I went for it and cleaned the function to leave only the ufunc=True implementation for newcalc=False, before pushing. That's why you didn't see it, it was a temporary version for benchmarking before pushing.

PR #309, opened 2019-12-02
Feature nstx

flothesof:
I've just added the geometry for NSTX based on https://iopscience.iop.org/article/10.1088/0029-5515/52/5/054012/meta
Adds the geometry as well as an entry in the table of configurations.


lasofivec:
Do you think we could add the obtained 2D cam image as an example ?

Didou09:
We probably should,
At some point, when we have a bit more tokamaks, we'll probably plot all of them in a common figure to give an idea of their respective scale and with their associated plot_touch() images, to give a nice overview of all of them

Issue #310, opened 2019-12-03
Creating camera fails 


flothesof:
When I run the following example code from devel 2d9b1837a126d4ad9f88b8b7779ab5ddd58a5f66:

```python
import matplotlib.pyplot
import numpy as np
import tofu as tf

config = tf.geom.utils.create_config()

cam2d = tf.geom.utils.create_CamLOS2D(
    config=config,
    P=[3.4, 0, 0],
    N12=100,
    F=0.1,
    D12=0.1,
    angs=[np.pi, np.pi/6, 0],
    Name="",
    Exp="",
    Diag="",
)
```

I get the following error:
``` 
  File "C:\Users\FL232714\Documents\GitHub\tofu\tofu\geom\utils.py", line 1011, in create_CamLOS2D
    return _create_CamLOS(nD=2, **kwdargs)
  File "C:\Users\FL232714\Documents\GitHub\tofu\tofu\geom\utils.py", line 1000, in _create_CamLOS
    color=color, config=config, SavePath=SavePath)
  File "C:\Users\FL232714\Documents\GitHub\tofu\tofu\geom\_core.py", line 3607, in __init__
    super(Rays, self).__init__(**kwdargs)
  File "C:\Users\FL232714\Documents\GitHub\tofu\tofu\utils.py", line 1388, in __init__
    self._init(**kwdargs)
  File "C:\Users\FL232714\Documents\GitHub\tofu\tofu\geom\_core.py", line 4033, in _init
    self._set_dgeom(sino=True, **kwdgeom)
  File "C:\Users\FL232714\Documents\GitHub\tofu\tofu\geom\_core.py", line 4588, in _set_dgeom
    self.compute_dgeom(extra=extra)
  File "C:\Users\FL232714\Documents\GitHub\tofu\tofu\geom\_core.py", line 4475, in compute_dgeom
    Nstep=50,
  File "C:\Users\FL232714\Documents\GitHub\tofu\tofu\geom\_plot.py", line 787, in _LOS_calc_InOutPolProj_Debug
    fs=fs, wintit=wintit, draw=False)
  File "C:\Users\FL232714\Documents\GitHub\tofu\tofu\geom\_core.py", line 3292, in plot
    Test=Test,
  File "C:\Users\FL232714\Documents\GitHub\tofu\tofu\geom\_plot.py", line 198, in Struct_plot
    draw=False, **dplot[proj])
[...]

PR #311, opened 2019-12-03
Issue310 more robust warning for out of vessel los

Didou09:
Main changes:
----------------
* Try / except introduced in tf.geom.Rays._compute_dgeom() in case where some LOS have no visibility inside vessel to make sure an error during the plotting process does not stop computation
* Exception is still caught and displayed as warning

Issues:
-------
* Fixes, in devel, issue #310

### DIFF
--- a/tofu/geom/_core.py
+++ b/tofu/geom/_core.py
@@ -4461,19 +4461,24 @@ class Rays(utils.ToFuObject):
             msg += repr(ind.nonzero()[0])
             warnings.warn(msg)
             if plotdebug:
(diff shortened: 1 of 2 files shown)


flothesof:
I've just found out where the problem is. Taking out line 167 in tofu/geom/_def.py makes the issue pass.
A good way to fix this could also be to add a unit test with a camera that has missing LOS so that our tests check that the debug plot works.

lasofivec:
Should we close this PR then ?

Didou09:
Yes, you're right, PR #313 will do, this one is no use anymore

PR #313, opened 2019-12-03
fix issue #310

flothesof:
Proposed changes instead of thos from #311.
This fixes the matplotlib plot error.
Also, I'd like to add a test that triggers the debug plot, but I'm unsure where to put it.
I've just browsed the test files in tofu/tests/tests01_geom but they look really messy... (sorry)
Where should I put my test for this?

### DIFF
--- a/tofu/geom/_core.py
+++ b/tofu/geom/_core.py
@@ -4430,10 +4430,16 @@ class Rays(utils.ToFuObject):
         indout[0, :] = indStruct[indout[0, :]]
         return kIn, kOut, vperp, indout, indStruct
 
-    def compute_dgeom(self, extra=True, plotdebug=True):
+    def compute_dgeom(self, extra=True, show_debug_plot=True):
+        """
+        Computes the dgeom attribute (what does dgeom mean? I'm guessing it stands for shortened dict_geometry?).
+
+        :param extra: bool, whether to perform some extra computations (which ones?)
+        :param show_debug_plot: bool, whether to show a debugging plot for invalid LOSes
+        """
         # Can only be computed if config if provided
         if self._dconfig["Config"] is None:
-            msg = "Attribute dgeom cannot be computed without a config !"
+            msg = "Attribute dgeom cannot be computed without a config!"
             warnings.warn(msg)
             return
 
@@ -4444,13 +4450,13 @@ class Rays(utils.ToFuObject):
         # Perform computation of kIn and kOut
         kIn, kOut, vperp, indout, indStruct = self._compute_kInOut()
 
-        # Clean up (in case of nans)
+        # Check for LOS that have no visibility inside the plasma domain
         ind = np.isnan(kIn)
         kIn[ind] = 0.0
         ind = np.isnan(kOut) | np.isinf(kOut)
         if np.any(ind):
             kOut[ind] = np.nan
-            msg = "Some LOS have no visibility inside the plasma domain !\n"
+            msg = "Some LOS have no visibility inside the plasma domain!\n"
             msg += "Nb. of LOS concerned: %s out of %s\n" % (
                 str(ind.sum()),
                 str(kOut.size),
@@ -4460,10 +4466,9 @@ class Rays(utils.ToFuObject):
             msg += "\nIndices of LOS with no visibility:\n"
             msg += repr(ind.nonzero()[0])
             warnings.warn(msg)
-            if plotdebug:
+            if show_debug_plot:
                 PIn = self.D[:, ind] + kIn[None, ind] * self.u[:, ind]
                 POut = self.D[:, ind] + kOut[None, ind] * self.u[:, ind]
-                # To be updated
                 _plot._LOS_calc_InOutPolProj_Debug(
                     self.config,
                     self.D[:, ind],
--- a/tofu/geom/_def.py
+++ b/tofu/geom/_def.py
@@ -164,7 +164,6 @@ def Plot_3D_plt_Tor_DefAxes(fs=None, wintit='tofu'):
     ax.set_xlabel(r"X (m)")
     ax.set_ylabel(r"Y (m)")
     ax.set_zlabel(r"Z (m)")
-    ax.set_aspect(aspect="equal", adjustable='datalim')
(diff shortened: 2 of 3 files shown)


flothesof:
What would you recommend for tests to make sure this doesn't break again? Which test file should it go in?

Didou09:
For the test, it should be in tofu/tests/tests01_geom/tests03_core.py
In the test class Test03_Rays(), you can add a method called

```
def testXX_shortdescription(self):
    pass 
```

tofu/geom/_def.py
```diff
@@ -164,7 +164,6 @@ def Plot_3D_plt_Tor_DefAxes(fs=None, wintit='tofu'):
     ax.set_xlabel(r"X (m)")
     ax.set_ylabel(r"Y (m)")
     ax.set_zlabel(r"Z (m)")
-    ax.set_aspect(aspect="equal", adjustable='datalim')
```
Didou09: Good idea,

However, I would recommend keeping the try / except, because 3d plotting in matplotlib is not very stable,
This plot, for instance, used to work fine until a matploltib update.
Given that 3d plotting is not very stable, I would recommend both solution:
* Try to make it work with your fix
* Keep the try / except as back-up safety in case it stops working for any reason at some point

Can you update your PR by adding the try / except with warning as in PR #311 ?

flothesof:
Sure, will do this as a next step!

tofu/geom/_core.py
```diff
@@ -4430,10 +4430,16 @@ def _compute_kInOut(self, largs=None, dkwd=None, indStruct=None):
         indout[0, :] = indStruct[indout[0, :]]
         return kIn, kOut, vperp, indout, indStruct
 
-    def compute_dgeom(self, extra=True, plotdebug=True):
+    def compute_dgeom(self, extra=True, show_debug_plot=True):
+        """
+        Computes the dgeom attribute (what does dgeom mean? I'm guessing it stands for shortened dict_geometry?).
```
lasofivec: I would say so too. I would let @Didou09 confirm :)
It looks like Florian wrote in the doc some questions, would you mind checking them out (also for next parameter)

tofu/geom/_def.py
lasofivec: I understand that `aspect="equal"` causes the error, but does the second parameter should also be removed?

flothesof:
After further thinking about the issue I've arrived at the conclusion that it's better to not add the try except block.

Supporting a debug plot is a feature that is very useful. It should always work from now on. 

The two proposed ways to act are:

- change 1: add a try/catch feature that prints an error message if something goes wrong with the debug plot
- change 2: add a test to the unit tests that makes sure the debug plot works

The worst-case scenario is that the debug plot breaks somehow in the future.
If this happens:

- change 1: a silent warning will be printed to the user who sees this, who might not report it as a bug.
- change 2: unit tests will break, prompting the maintainers to fix the bug.

Implementing change 1 and 2 is not useful because change 2 alone will make sure the debug plot works.

Hence my conclusion: change 2 is the way to go.

tofu/geom/_core.py
flothesof: Do you have any idea what sort of information dgeom contains?

tofu/geom/_def.py
flothesof: Good point, I'm going to try.

flothesof: No, it doesn't work.
```
ax.set_aspect(adjustable='datalim')
TypeError: set_aspect() missing 1 required positional argument: 'aspect'
```

lasofivec: Ok, it bothers me though to just take out this line. @Didou09 : i'm under the impression that it worked in some cases, no? I have already used this feature of tofu to debug a camera. We could try to replace it with a hack like: https://stackoverflow.com/questions/13685386/matplotlib-equal-unit-length-with-equal-aspect-ratio-z-axis-is-not-equal-to

Didou09:
Agreed, but:

* the bug may very well be matplotlib-version specific
* It will work for some users, and not for others, and we cant' force users to upgrade to a specific matplotlib version (too demanding, may break their environment)

=> if we go without the try / except, it means we may have to write tests on the current version of matplotlib to make sure the plot works fine for all versions (i.e.: extra workload).
We can go this way, but I want to be sure everyone is aware of this before we decide

flothesof:
> * the bug may very well be matplotlib-version specific
As far as I can tell, it is.

> * It will work for some users, and not for others, and we cant' force users to upgrade to a specific matplotlib version (too demanding, may break their environment)
I don't think it will work for some/not for others. What I am saying is: this is a matplotlib issue, for which we should only write "correct matplotlib code" and not care about the rest. If matplotlib works as expected this will work for everyone.

tofu/geom/_def.py
flothesof: It may have been a backwards incompatible change in matplotlib no?

Didou09:
> I don't think it will work for some/not for others. What I am saying is: this is a matplotlib issue, for which we should only write "correct matplotlib code" and not care about the rest. If matplotlib works as expected this will work for everyone.

Matplotlib will work as expected, except the API may change from one version to another. It is not necessarily a matplotlib bug, more a change of API.
So it will work for those with the good version / API, not for the others, since everyone does not have the same version.
From my point of view, the question is more "If it does not work because of matploltib what do we do?":
* Implement a try / except that will make sure it never blocks the computation whatever the API
 * Implement version / API-specific tests ?

flothesof:
I would say that if we notice a problem, which I think is not super likely, we adjust the required version of matplotlib in the setup.
This allows us to have control over what works while not adding extra complexity.

Didou09:
This issue was finally adressed in a separate PR  #343 .
After some updates on our system here at IRFM, it appears that 3D plottin is definitely not stable / clean with matplotlib and seems to crash in several versions.
Assuming the 3d debug plot will work is too strong an hypothesis.
I decided to remove this 3D plot until either matplotlib gets more stable 3D plot or until we include support for PyQtGraph. 
In the meantime, I included a more informative error message.

Issue #318, opened 2019-12-06
method for easy adding of adding default ids for synthetic diag to MultiIDSLoader


Didou09:
For tabulated default ids-es, it should be easy to add them once the diag ids is included

Didou09:
Fixed, in devel by PR #319

PR #319, opened 2019-12-09
Issue318 add synth diag ids

Didou09:
Motivations:
--------------
* For non-experienced users, the imas interface class MultiIDSLoader shall provide an easy way to:
     - load diagnostics geometries
     - load experimental diagnostic data
     - load all ids necessary to compute synthetic signal for a given diagnostic
* The current status allowed to do that in ~5-6 lines, provided the user knows where to get the info
* This need being a typical use-case, it should be easier / faster

Main changes:
----------------
* All diagnostocs for which a synthetic diagnostic exists are tabulated inside the class hidden attributes, along with their corresponding ids (core_profiles...)
* A method allows to instantaneously get the ids necessary for synthetic data (get_input_for_synthdiag())
* A method, add_ids_synthdiag(), calls the previous to automatically add and load then necessary ids, if necessary with a custom idd defined just-in-time.
* Two keywords added at instanciation (ids_base and synthdiag) for users who want to load everythong at once from a single idd)
* (extra) a get_events() method is implemented to easily print the chronologial events in a shot from ids 'pulse_schedule' 

Issues:
-------
* Fixes, in devel, issue #318 

Example (with a custom idd):
--------------------------------

**Before (5 lines, a bit complicated)**
```
> import tofu as tf

# load experimental data and wall for geometry, and pulse_schedule for tIGNITRON
> multi = tf.imas2tofu.MultiIDSLoader(shot=53259, ids=['pulse_schedule', 'wall', 'interferometer'], get=False)

# Find which ids are necessary for synthetic signal calculation
> print(multi._didsdiag['interferometer']['synth']['dsig'])
{'core_profiles': ['t', '1dne', '1drhotn'],
 'equilibrium': ['t', '2drhotn', '2dmeshNodes', '2dmeshFaces']}

# add these ids to multi
> multi.add_ids(ids=['core_profiles', 'equilibrium'], shot=53259, user='CB165101', run=1, get=False)

# Load everything that was added
> multi.open_get_close()

# Calculate signal
> sig = multi.calc_signal('interferometer')
```

**Now (3 lines, simpler)**
```
> import tofu as tf

# load experimental data (wall and pulse_schedule are automatically included)
> multi = tf.imas2tofu.MultiIDSLoader(shot=53259, ids='interferometer', get=False)

# Add ids necessary for calculation (automatically identified: tabulated in class attribute)
> multi.add_ids_synth_diag(shot=53259,  user='CB165101', run=1)

# Calculate signal
> sig = multi.calc_signal('interferometer')
```

Also, events can now be obtained using (will only work if 'pulse_schedule' was properly populated):
`> multi.get_events()`

### DIFF
--- a/tofu/imas2tofu/_core.py
+++ b/tofu/imas2tofu/_core.py
@@ -614,7 +614,7 @@ class MultiIDSLoader(object):
                  'core_profiles':['t','Te','ne']}
                }
 
-
+    _IDS_BASE = ['wall', 'pulse_schedule']
 
 
     ###################################
@@ -625,7 +625,8 @@ class MultiIDSLoader(object):
 
     def __init__(self, preset=None, dids=None, ids=None, occ=None, idd=None,
                  shot=None, run=None, refshot=None, refrun=None,
-                 user=None, tokamak=None, version=None, get=None, ref=True):
+                 user=None, tokamak=None, version=None,
+                 ids_base=None, synthdiag=None, get=None, ref=True):
         super(MultiIDSLoader, self).__init__()
 
         # Initialize dicts
@@ -642,6 +643,14 @@ class MultiIDSLoader(object):
             self.add_ids(preset=preset, ids=ids, occ=occ, idd=idd, get=False)
             if get is None and (ids is not None or preset is not None):
                 get = True
+            if ids_base is None:
+                ids_base = True
+            if ids_base is True:
+                self.add_ids_base(get=False)
+            if synthdiag is None:
+                synthdiag = False
+            if synthdiag is True:
+                self.add_ids_synthdiag(get=False)
         else:
             self.set_dids(dids)
             if get is None:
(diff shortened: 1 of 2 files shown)


tofu/imas2tofu/_core.py
```diff
@@ -625,7 +625,8 @@ def _rhosign(rho, theta):
 
     def __init__(self, preset=None, dids=None, ids=None, occ=None, idd=None,
                  shot=None, run=None, refshot=None, refrun=None,
-                 user=None, tokamak=None, version=None, get=None, ref=True):
+                 user=None, tokamak=None, version=None,
+                 ids_base=None, synthdiag=None, get=None, ref=True):
```
lasofivec: For future devs (and yourself), it could be a nice to add a bit of doc

Didou09: Good point, will add docstr in future PR
Thanks !

Issue #301, opened 2019-11-28
accessing private method in tutorial


lasofivec:
Normally methods/parameters starting with "_" are private.
However in a tutorial we do:

`tf.geom.utils._DCONFIG_TABLE.keys`

it looks to me like there should be a function

`tf.geom.utils.get_all_configs_key()`

for example.

Didou09:
Yes definitely,

Will do it in the coming days

Didou09:
Fixed, in devel, by PR #327

PR #327, opened 2020-01-13
#301 List of available config accessible for end-users

Didou09:
Motivations:
--------------
As pointed out by issue #301 , in the [online tutorial for plotting the gallery of fusion machines](https://tofuproject.github.io/tofu/auto_examples/tutorials/tuto_plot_gallery_fusion_machines.html), we ask the user to read a hidden variable `tf.geom.utils._DCONFIG_TABLE`. 

This is not good practice, a public method should exist to list all available pre-implemenetd configurations.

Main changes:
----------------
Two new methods are created, a hidden and a public:
* `tf.geom.utils._get_listconfig()` : a hidden method used to generate the dict of (unique keys, shortcuts) from hidden variables `_DCONFIG` and `_DCONFIG_TABLE` and return either a fully formatted str message or the dict itself. It is used to provide the core message of the public method, and also re-used to properly generate the error message when `tf.geom.utils.create_Config()` is not properly called by the user.
* `tf.geom.utils.get_available_config()`: a public method that prints a fully formatted explicit message containing all available keys to pre-implemented config

Example:
----------
```
In [1]: import tofu as tf

In [2]: tf.geom.utils.get_available_config()
A config is the geometry of a tokamak
You can define your own, see online tutorial at:
    https://tofuproject.github.io/tofu/auto_examples/tutorials/tuto_plot_create_geometry.html
tofu also also provides some pre-defined config ready to load
They are available via their name or via shortcuts

	unique names	shortcuts
	------------ 	-------------------------
	- ITER-V1     	['ITER-V1', 'A2']
	- ITER-V2     	['ITER-V2', 'B4', 'ITER']
	- JET-V0      	['JET-V0', 'JET']
	- NSTX-V0     	['NSTX-V0', 'NSTX']
	- WEST-Sep    	['WEST-Sep', 'A3']
	- WEST-V1     	['WEST-V1', 'A1']
	- WEST-V2     	['WEST-V2', 'B1']
	- WEST-V3     	['WEST-V3', 'B2']
	- WEST-V4     	['WEST-V4', 'B3', 'WEST']

  => to get a pre-defined config, call for example:
	config = tf.geom.utils.create_config('ITER')
```

Issues:
-------
Fixes, in devel, issue #301


tofu/geom/utils.py
```diff
@@ -695,8 +700,40 @@ def _compute_CamLOS2D_pinhole(P=None, F=0.1, D12=0.1, N12=100,
                   'B4': 'ITER-V2',
                   'NSTX': 'NSTX-V0'}
 
-# Default config
-_DEFCONFIG = 'ITER'
+
+def _get_listconfig(dconfig=_DCONFIG, dconfig_table=_DCONFIG_TABLE,
```
Didou09: core hidden method
It is the base that provides either a dict or a str (formatted message) containing the core info

tofu/geom/utils.py
```diff
@@ -695,8 +700,40 @@ def _compute_CamLOS2D_pinhole(P=None, F=0.1, D12=0.1, N12=100,
...
+        l1 = np.max([len(str(v0)) for v0 in dc.values()] + [len('shortcuts')])
+        msg = ("\n\t" + "unique names".ljust(l0) + "\tshortcuts"
+               + "\n\t" + "-"*l0 + " \t" + "-"*l1
+               + "\n\t- "
+               + "\n\t- ".join(["{}\t{}".format(k0.ljust(l0), v0)
+                                for k0, v0 in dc.items()]))
+        return msg
+
+
+def get_available_config(dconfig=_DCONFIG, dconfig_table=_DCONFIG_TABLE,
```
Didou09: Public method for end-users

tofu/geom/utils.py
```diff
@@ -724,11 +761,9 @@ def _create_config_testcase(config=None, returnas='object',
         config = dconfig_table[config]
 
     else:
-        msg = ("The provided config name is not valid:\n"
-               + "Please choose among either:\n"
-               + "\t - unique keys: {}\n".format(list(dconfig.keys()))
-               + "\t - shortcuts  : {}\n\n".format(list(dconfig_table.keys()))
-               + "  => you provided: case = {}\n".format(config))
+        msg = ("\nThe provided config name is not valid.\n"
```
Didou09: This error message re-uses the formatted message created by the hidden method

tofu/geom/utils.py
flothesof: This is great. However, I would suggest adding a short docstring to the new method and its intended use.
In particular, this would allow to describe in words what _DCONFIG and _DCONFIG_TABLE is
(to me, it's not clear what the difference is).

tofu/geom/utils.py
flothesof: Same here, a docstring would be nice.

Also, since this is a new public method, it should be added to `__all__` no? As far as I can tell this is not the case.

Otherwise the message is nicely written, especially the use of an URL is good I believe.

tofu/geom/utils.py
Didou09: Done

tofu/geom/utils.py
Didou09: Done, thanks :-)

tofu/geom/utils.py
```diff
@@ -680,28 +688,78 @@ def _compute_CamLOS2D_pinhole(P=None, F=0.1, D12=0.1, N12=100,
...
+               + "\n\t- ".join(["{}\t{}".format(k0.ljust(l0), v0)
+                                for k0, v0 in dc.items()]))
+        return msg
 
-# Default config
-_DEFCONFIG = 'ITER'
+
+def get_available_config(dconfig=_DCONFIG,
+                         dconfig_shortcuts=_DCONFIG_SHORTCUTS,
+                         verb=True, returnas=False):
```
flothesof: Why is returnas a bool here? 
Below there's a check `if returnas is str`... Something looks weird here.

flothesof: Why is returnas a bool here? Shouldn't it be str? 
Below there's a check `if returnas is str`... 
What do you think?

Didou09: Nope, in general in many tofu functions / methods, returnas can be:
* False => nothing is returned
* None / True => set to default return format
* a specific format (here str)

It is because in this function there is only one output possible format, but that may evolve in the future.
The use of True / None is to specify we want output but either the user doesn't know that he has choice between several formats (in that case it is usually set to the most intuitive) or to let the format be decided later by another function.

I agree in this case we just have a choice between False (default) and str, but this way we keep the general workflow / convention

Didou09: I did not implement True / None because I forgot though,
But in 99 % cases users will not ask for returning anything

Issue #331, opened 2020-01-21
_GG.Poly_isClockwise not working for some polygons


Didou09:
Trying to add the new geometry of ASDEX Upgrade (AUG), that a user sent to us, I ran into an issue that seems to be due to `_GG.Poly_isClockwise()`

I just updated branch **_Issue281_NewDefaultConfigs_** (merged with devel and pushed to github).
I also re-compiled the cython binaries.

But when trying to create the AUG config from the pre-defined txt files (containing the R, Z polygons of each structure), 3 polygons are not handled properly (the config is still created because I implemented a try/except on each structure):

```
In [1]: import tofu as tf
conf = /Home/DV226270/ToFu_All/tofu_git/tofu/tofu/imas2tofu/__init__.py:83: UserWarning: 
You do not seem to be using the latest IMAS version:
'module list' vs 'module av IMAS' suggests:
	- Current version: 3.23.1-4.0.3
	- Latest version : 3.25.0-4.4.0
  warnings.warn(msg)
/Home/DV226270/ToFu_All/tofu_git/tofu/tofu/__init__.py:95: UserWarning: 
The following subpackages are not available:
    - tofu.mag
  => see print(tofu.dsub[<subpackage>]) for details.
  warnings.warn(msg)

In [2]: conf = tf.geom.utils.create_config('AUG')
/Home/DV226270/ToFu_All/tofu_git/tofu/tofu/geom/utils.py:838: UserWarning: Could not be loaded: TFG_PFC_ExpAUG_D2dBu1.txt
  warnings.warn(msg)
/Home/DV226270/ToFu_All/tofu_git/tofu/tofu/geom/utils.py:838: UserWarning: Could not be loaded: TFG_PFC_ExpAUG_D2dBu2.txt
  warnings.warn(msg)
/Home/DV226270/ToFu_All/tofu_git/tofu/tofu/geom/utils.py:838: UserWarning: Could not be loaded: TFG_PFC_ExpAUG_SBi.txt
  warnings.warn(msg)

In [3]: conf.plot()
```

For every structure failing, there is warning. If you plot the config you'll notice it is missing some grey structures in the right hand side of the poloidal cross-section:

![issues331](https://user-images.githubusercontent.com/5929562/72824834-3b7bd480-3c76-11ea-848e-4e1c0165dbdb.png)

Could this have something to do with the bug corrected in PR #323 ?
(the txt files themselves are in tofu/geom/inputs)

lasofivec:
[...]

PR #332, opened 2020-01-21
[bf #331] raising exceptions when poly ill defined

lasofivec:
As discussed in #331 : 
The polygon needs to be properly defined in order for the algorithm `isClockwise` works correctly.
If it finds an issue (null volume) it will raise an exception.

The exception is caught and raised whenever function used
**Warning**: I didn't manage to properly catch and raise the error.  Can you see what I am doing wrong ? I left a `print` for the moment

I also realized in _core, `isClockwise` is defined sometimes **before** checking if the polygon is closed, however the algorithm needs for the polygon to be closed.

Lastly: some flake8 changes

### DIFF
--- a/tofu/geom/_GG.pyx
+++ b/tofu/geom/_GG.pyx
@@ -207,7 +207,7 @@ def Poly_isClockwise(np.ndarray[double,ndim=2] Poly):
     cdef int idp1 = (idmin + 1) % npts
     cdef str err_msg = ""
     if idmin == 0 :
-        idm1 = npts - 1
+        idm1 = npts - 2
     res = mvx[idm1]  * (mvy[idmin] - mvy[idp1]) + \
           mvx[idmin] * (mvy[idp1]  - mvy[idm1]) + \
           mvx[idp1]  * (mvy[idm1]  - mvy[idmin])
--- a/tofu/geom/_GG.pyx
+++ b/tofu/geom/_GG.pyx
@@ -200,14 +200,27 @@ def Poly_isClockwise(np.ndarray[double,ndim=2] Poly):
     cdef double res
     cdef double[:,::1] mv_poly = np.ascontiguousarray(Poly)
     cdef int npts = mv_poly.shape[1]
-    cdef double[::1] mvx = mv_poly[0,:]
-    cdef double[::1] mvy = mv_poly[1,:]
-    cdef int idmin = _bgt.find_ind_lowerright_corner(mvx, mvy, npts)
-    cdef int idm1 = idmin - 1
-    cdef int idp1 = (idmin + 1) % npts
+    cdef int ndim = mv_poly.shape[0]
+    cdef double[::1] mvx
+    cdef double[::1] mvy
+    cdef int idmin
+    cdef int idm1
+    cdef int idp1
     cdef str err_msg = ""
+    # Checking that Poly wasn't given in the shape (npts, ndim)
+    if ndim > npts:
+        mv_poly = np.ascontiguousarray(Poly.T)
+        npts = mv_poly.shape[1]
+        ndim = mv_poly.shape[0]
+    mvx = mv_poly[0,:]
+    mvy = mv_poly[1,:]
+    # Getting index of lower right corner and its neighbors
+    idmin = _bgt.find_ind_lowerright_corner(mvx, mvy, npts)
+    idm1 = idmin - 1
+    idp1 = (idmin + 1) % npts
     if idmin == 0 :
         idm1 = npts - 2
+    # Computing area of lower right triangle
     res = mvx[idm1]  * (mvy[idmin] - mvy[idp1]) + \
           mvx[idmin] * (mvy[idp1]  - mvy[idm1]) + \
           mvx[idp1]  * (mvy[idm1]  - mvy[idmin])
@@ -219,7 +232,6 @@ def Poly_isClockwise(np.ndarray[double,ndim=2] Poly):
                     + ", " + str(mvy[idmin]) + ".\n"
                     + "   The two neighboring points are : "
                     + str(idm1) + " and " + str(idp1) + ".")
-        print(err_msg)
         raise Exception(err_msg) # not working
     return res < 0.
 
--- a/tofu/geom/_GG.pyx
+++ b/tofu/geom/_GG.pyx
@@ -188,16 +188,14 @@ def CoordShift(Pts, In='(X,Y,Z)', Out='(R,Z)', CrossRef=None):
 
 def Poly_isClockwise(np.ndarray[double,ndim=2] Poly):
     """ Assuming 2D closed Poly !
-    TODO @LM :
     http://www.faqs.org/faqs/graphics/algorithms-faq/
-    A slightly faster method is based on the observation that it isn't
-    necessary to compute the area.  Find the lowest vertex (or, if
-    there is more than one vertex with the same lowest coordinate,
-    the rightmost of those vertices) and then take the cross product
-    of the edges fore and aft of it.  Both methods are O(n) for n vertices,
-    but it does seem a waste to add up the total area when a single cross
-    product (of just the right edges) suffices.  Code for this is
-    available at ftp://cs.smith.edu/pub/code/polyorient.C (2K).
+    Find the lowest vertex (or, if there is more than one vertex with
+    the same lowest coordinate, the rightmost of those vertices) and then
+    take the cross product of the edges before and after it.
+    Both methods are O(n) for n vertices, but it does seem a waste to add up
+    the total area when a single cross product (of just the right edges)
+    suffices.  Code for this is available at
+    ftp://cs.smith.edu/pub/code/polyorient.C (2K).
     """
     cdef double res
     cdef double[:,::1] mv_poly = np.ascontiguousarray(Poly)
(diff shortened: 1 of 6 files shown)


tofu/geom/_GG.pyx
```diff
@@ -206,12 +204,23 @@ def Poly_isClockwise(np.ndarray[double,ndim=2] Poly):
...
           mvx[idp1]  * (mvy[idm1]  - mvy[idmin])
+    if abs(res) < _VSMALL:
+        err_msg += ("In Poly_isClockwise : \n"
+                    + "   Found lowest right point at index : "
+                    + str(idmin)
+                    + ", of coordinates :" + str(mvx[idmin])
+                    + ", " + str(mvy[idmin]) + ".\n"
+                    + "   The two neighboring points are : "
+                    + str(idm1) + " and " + str(idp1) + ".")
+        print(err_msg)
```
lasofivec: added this since I didnt manage to catch the Exception later

tofu/geom/_GG.pyx
```diff
@@ -274,8 +283,11 @@ def Poly_Order(np.ndarray[double,ndim=2] Poly, str order='C', Clock=False,
     if not np.allclose(poly[:,0],poly[:,-1], atol=_VSMALL):
         poly = np.concatenate((poly,poly[:,0:1]),axis=1)
     if poly.shape[0]==2 and not Clock is None:
-        if not Clock==Poly_isClockwise(poly):
-            poly = poly[:,::-1]
+        try:
+            if not Clock==Poly_isClockwise(poly):
+                poly = poly[:,::-1]
+        except Exception as excp:
+            raise excp
```
lasofivec: This is where is called when creating the config. Tried several things...

tofu/geom/_comp.py
```diff
@@ -34,9 +34,12 @@ def _Struct_set_Poly(
...
     # Make Poly closed, counter-clockwise, with '(cc,N)' layout and arrayorder
-    Poly = _GG.Poly_Order(
-        Poly, order="C", Clock=False, close=True, layout="(cc,N)", Test=True
-    )
+    try:
+        Poly = _GG.Poly_Order(
+            Poly, order="C", Clock=False, close=True, layout="(cc,N)", Test=True
+            )
+    except Exception as excp:
+        print(excp)
```
lasofivec: and trying to catch and print the exception message...

tofu/geom/_core.py
```diff
@@ -5574,17 +5574,17 @@ def _kInOut_Isoflux_inputs_usr(self, lPoly, lVIn=None):
         if type(lPoly) is list:
             for ii in range(nPoly):
                 # Check closed and anti-clockwise
-                if _GG.Poly_isClockwise(lPoly[ii]):
-                    lPoly[ii] = lPoly[ii][:, ::-1]
                 if not np.allclose(lPoly[ii][:, 0], lPoly[ii][:, -1]):
                     lPoly[ii] = np.concatenate(
                         (lPoly[ii], lPoly[ii][:, 0:1]), axis=-1
                     )
```
lasofivec: first we close the poly then we check if clockwise

tofu/geom/_core.py
```diff
@@ -6900,7 +6906,7 @@ def get_summary(
         # Prepare
         kout = self._dgeom["kOut"]
         indout = self._dgeom["indout"]
-        lS = self._dconfig["Config"].lStruct
+        # lS = self._dconfig["Config"].lStruct
```
lasofivec: is not being used, so I commented it

tofu/geom/_basic_geom_tools.pxd
```diff
@@ -1,6 +1,7 @@
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: cdivision=True
+# cython: initializedcheck=False
```
Didou09: What does this flag do ?

lasofivec:
Ok, my bad. Forgot that it needed to be a "-2" instead of "-1". 
This solved the issue, event hough I am not entirely sure why did this also solve the issue with the plotting of the structure ?

Capture d’écran 2020-01-22 à 13 57 00

tofu/geom/_basic_geom_tools.pxd
lasofivec: Cython normally checks if a memory view is initialized before using it. This turns the verification off (for speed purposes). I wanted to match the flags with the GG file.

tofu/geom/_GG.pyx
```diff
@@ -206,12 +204,23 @@ def Poly_isClockwise(np.ndarray[double,ndim=2] Poly):
...
+    if abs(res) < _VSMALL:
+        err_msg += ("In Poly_isClockwise : \n"
+                    + "   Found lowest right point at index : "
+                    + str(idmin)
+                    + ", of coordinates :" + str(mvx[idmin])
+                    + ", " + str(mvy[idmin]) + ".\n"
+                    + "   The two neighboring points are : "
+                    + str(idm1) + " and " + str(idp1) + ".")
+        print(err_msg)
+        raise Exception(err_msg) # not working
```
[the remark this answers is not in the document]
lasofivec: it is being raised if the function is called directly called from python. For example, the test that is failing is showing:
``

Didou09:
It's because the plotting procedure also depends on the order of the polygon (clockwise or not).
The position of the labels (indices) is determined by the in/out vectors perpendicular to each segment, and these in/out vectors may be wrong if there was a bug in one of the preceding steps.

Didou09:
I'm testing it locally, but it seems we are fine,
I just want to double-check because of the C-contiguous error I found last time

Didou09:
So, locally it seems we are good,
But travis finds an error in the unit tests:
```
======================================================================
ERROR: tofu.tests.tests01_geom.tests01_GG.test02_Poly_CLockOrder
----------------------------------------------------------------------
  File "/home/travis/miniconda/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/ToFuProject/tofu/tofu/tests/tests01_geom/tests01_GG.py", line 119, in test02_Poly_CLockOrder
    not GG.Poly_isClockwise(P), P.flags['C_CONTIGUOUS'],
  File "tofu/geom/_GG.pyx", line 223, in tofu.geom._GG.Poly_isClockwise
Exception: In Poly_isClockwise : 
   Found lowest right point at index : 1, of coordinates :0.0, 0.0.
   The two neighboring points are : 0 and 0.
```

Could you have a look ?

lasofivec:
I am hunting this bug :)

lasofivec:
Ok it took me a while to understand the bug but it is not really a bug.....
This is what happens:

- A polygon is created in the right order: (number_of_dims, number_of_points) = (2, npts)
- It is passed to the `Poly_Order` function with the parameter `order=(N,cc)` meaning that the polygon is going to be translated so that its shape is (npts, 2)
- Then we check if the polygon is not clockwise, however in this algorithm we suppose that the polygon is in the order (2, npts).
- Original version of the algorithm (coded by you, see below) and the version in devel, the computation was done and because the algorithm will return False since it's being computed on a segment instead of the polygon. 
- Now we raise an exception if it finds that the volume is 0 (which is the case here), thus the assert `assert not GG.Poly_isClockwise(P)` fails...

So, I would suggest either: 
- Taking out the assert in the test function and maybe adding a test to check if the number of points in `GG.Poly_isClockwise` is > 2.
- Checking if we find npts < 3, and in that case translate the polygon, and let the algorithm run as is. If for example we send a polygon with shape (2,2), it will still fail (I can add in the exception message the shape of the polygon received).

I wouldn't suggest calling `GG.Poly_Order` in `GG.Poly_isClockwise` since `GG.Poly_Order` already calls `GG.Poly_isClockwise`.

FTR, original version of the function

```
@cython.cdivision(True)
@cython.wraparound(False)
@cython.boundscheck(False)
def Poly_isClockwise(cnp.ndarray[double,ndim=2] Poly):
    """ Assuming 2D closed Poly ! """
    cdef int ii, NP=Poly.shape[1]
    cdef double Sum=0.
    for ii in range(0,NP-1):
        Sum += Poly[0,ii]*Poly[1,ii+1]-Poly[0,ii+1]*Poly[1,ii]
    return Sum < 0.
 

```

lasofivec:
I think the second solution is the most elegant, so I made changes in that direction. Let me know what you think

Didou09:
Hum, I think I was originally translating from (2, npts) to (npts, 2) because before I was using external library Polygon2 (or Polygon3) to computed some properies (area...).
Since your optimizations, we got rid of this dependency, so we can probably stop translating from (2, npts) to (npts, 2) and stick with (2, npts) all the time.

hat should solve the issue, wouldn't it ?

Didou09:
Also, I agree your solution is the most elegant

lasofivec:
Yes, but this would mean we should make some major changes to PolyOrder:
- Take out the option of changing the order
- Renaming the function :)
- Checking all tests and code that call it

lasofivec:
Probably we can open a new issue about this if you prefer

Didou09:
Hum, good point, 

What do you think ?
I'll let you decide which option you prefer, either a hotfix (your option 2) or a full-blown rewriting of the incriminated parts.

Given that we are a bit late on time I would probably go for a hotfix, but I understand if you prefer something cleaner and more complete. your move ;-)

lasofivec:
we can go for the hotfix (in this branch and later in devel) and open an issue for the cleaner solution.

Issue #281, opened 2019-11-26
Add new default configs


Didou09:
Start implementing new configurations from other tokamaks, e.g.: AUG, COMPASS, EAST, JET, JT60-SA, KSTAR, MAST, TCV and other toroidal or linear devices

Didou09:
A first version of JET was added by PR #282

PR #333, opened 2020-01-23
Issue281 new default configs - AUG

Didou09:
Main changes:
----------------
* A bug was fixed in _GG.Poly_isClockwise() via PR #332 (fork of this branch)
* A new pre-defined tokamak geometry is available in an axisymmetric version: ASDEX Upgrade (AUG)

And AUG now appears in the available configs message created by PR #327 

```
In [1]: import tofu as tf 

In [2]: tf.geom.utils.get_available_config()
A config is the geometry of a tokamak
You can define your own, see online tutorial at:
	https://tofuproject.github.io/tofu/auto_examples/tutorials/tuto_plot_create_geometry.html
tofu also also provides some pre-defined config ready to load
They are available via their name or via shortcuts

	unique names	shortcuts
	------------ 	-------------------------
	- AUG-V1      	['AUG-V1', 'AUG']
	- ITER-V1     	['ITER-V1', 'A2']
	- ITER-V2     	['ITER-V2', 'B4', 'ITER']
	- JET-V0      	['JET-V0', 'JET']
	- NSTX-V0     	['NSTX-V0', 'NSTX']
	- WEST-Sep    	['WEST-Sep', 'A3']
	- WEST-V1     	['WEST-V1', 'A1']
	- WEST-V2     	['WEST-V2', 'B1']
	- WEST-V3     	['WEST-V3', 'B2']
	- WEST-V4     	['WEST-V4', 'B3', 'WEST']

  => to get a pre-defined config, call for example:
	config = tf.geom.utils.create_config('ITER')

In [3]: conf = tf.geom.utils.create_config('AUG')

In [4]: conf.plot()
```

PR_AUG

Perspectives:
--------------
* A more detailed version (with non-axisymmetric details for PFC) should be made available in a future version

Issues:
-------
* Progress on issue #281 

Acknowledgements:
---------------------
A contribution obtained with the help of @jdhare


tofu/geom/_core.py
```diff
@@ -436,19 +436,23 @@ def _checkformat_inputs_dgeom(
             Poly = Poly.T
 
         # Elimininate any double identical point
-        ind = np.sum(np.diff(Poly, axis=1) ** 2, axis=0) < 1.0e-12
+        ind = np.sum(np.diff(np.concatenate((Poly, Poly[:, 0:1]), axis=1),
```
lasofivec: why do you close the poly ?

tofu/geom/utils.py
```diff
@@ -795,20 +807,36 @@ def _create_config_testcase(config=None, returnas='object',
...
-            pfe = os.path.join(path,ff[0])
-            obj = eval('_core.'+cc).from_txt(pfe, Name=ss, Type='Tor',
-                                             Exp=dconfig[config]['Exp'],
-                                             out=returnas)
-            if returnas not in ['object', object]:
-                obj = ((ss,{'Poly':obj[0], 'pos':obj[1], 'extent':obj[2]}),)
-            lS.append(obj)
+            elif len(ff) > 1:
+                # More demanding criterion
+                ssbis, Expbis = '_'+ss+'.txt', '_Exp'+Exp+'_'
```
lasofivec: what are ss files ?

tofu/geom/_core.py
Didou09: We want a bool array ind the same shape as Poly, to tell us which successive points are redundant.
So:
* We use np.diff() => ind will be of shape (npts-), if poly has shape (2, npts). Concatenating the first poitn at the end (i.e. closing the polygon) will yield the good shape for ind
* If we don't close the poly here, and if the first point and the last point are identical (poly already closed), np.diff() will not detect it. So closing the poly in np.diff() is actually a good way of making sure we remove (at line 443) the last point if the poly was already closed

tofu/geom/utils.py
Didou09: In this loop, cc is the class str ('PFC', 'Ves'...) and ss is the name ('').
The name of a .txt file containing the polygon of a Struct is formatted as

TFG_class_ExpTokamak_Name.txt

For example:

TFG_PFC_ExpAUG_TPLT1.txt

This file contains the polygon for a PFC of experiment ASDEX Upgrade (AUG) which is called TPLT1.

Didou09: In this method we try to identify a unique txt fle corresponding to a given Struct (by class, Exp, name).
A first try is done with simple criteria which are sufficient 90% of the time.
But in some case multiple matches are found, because for example the names of several Struct share a common part, for example:

TFG_PFC_ExpAUG_D3BG1.txt
TFG_PFC_ExpAUG_D3BG10.txt

These are two different Struct. But the name of the first one is fully enclosed in the name of the second.
So if we look for all files containing the str 'D3BG1', we'll find these two solutions.
In that case, we search again with more strict criteria to identify the unique file.

Didou09:
Thanks for approving the merge :-)
I'm restoring the branch because this is branch that will live a long time since we will use it every time we'll add a new tokamak (I'm expecting GOLEM, TCV and JET in the coming weeks / months)

lasofivec:
You can restore it when it is needed but you can also keep it on :)

Issue #344, opened 2020-02-06
Export bin scripts via setup.py


Didou09:
tofu typically delivers at least 2 bash scripts to be executable driectly from the command line:
* **_tofuplot_** : load data from imas and plot it in an interactive figure
* **_tofucalc_** : load data from imas and use it to calculate synthetic data for a diagnostic and plot it in an interactive figure

They can be used only when the sub-package **_imas2tofu_** is available (i.e.: only when imas is available), but in that case they greatly simplify calls to tofu for users not familiar with python. They do not allow as many options as calling tofu from the ipython console, but they provide the basics for an everyday use.

These 2 binaries shall be automatically deloyed by setup.py at tofu installation in order to be available straight from the command line.

We'll use the [tool ](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html) that @lasofivec  found recently.

Most online sources mention that the method `entry_points `is preferrable over `scripts `because `scripts `is the old way, and cannot be easily tested, and `entry_points `allo for better portability and cross-platform compatibility via **_pip_**.

See [here](https://stackoverflow.com/questions/18787036/difference-between-entry-points-console-scripts-and-scripts-in-setup-py), and [here](https://chriswarrick.com/blog/2014/09/15/python-apps-the-right-way-entry_points-and-scripts/) for example

Didou09:
Fixed, in devel, by PR #345

lasofivec:
I changed the name to be a bit more explicit

PR #345, opened 2020-02-06
Issue344 deploybin

Didou09:
Main changes:
----------------
* automated deployment of **_tofuplot_** and **_tofucalc_** at `setup.py install` time thanks to the use of `entry_points`. both scripts were moved to `tofu/scripts/`
* Source: 
https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html
https://chriswarrick.com/blog/2014/09/15/python-apps-the-right-way-entry_points-and-scripts/
https://stackoverflow.com/questions/774824/explain-python-entry-points

Issues:
--------
Fixes, in devel, issue #344


Didou09:
Youu're absolutely right on every point:
* It does include the changes from PR #343 
* I should have started from devel

But in this particular case, I needed these changes to fix this issue (there cumulative so to speak), so I had not choice  ;-)
Thanks !

Issue #340, opened 2020-01-31
Anaconda deployment failing


Didou09:
Main issue:
------------

Version 1.4.2 could not be deployed to anaconda on the [tagged build](https://travis-ci.org/ToFuProject/tofu/builds/644286200)

The script failed with exit status 127
I don't know what it corresponds to but I notice that at least 1 environment variable used in the anaconda_upload.sh script is not defined: **$PKG_DIR**

In previous versions (1.4.1) it was defined in **_install_** section as:
`export PKG_DIR=$HOME/miniconda/conda-bld/linux-64/`

I suspect this is the cause of the issue.

Other remarks:
-----------------
* Looking at the build log, it seems that the section _**before_deploy**_ is run not just once but once before each deployment.
=> As a result the  `conda build conda_recipe` is run everytime, even for pip sdist deployment, this takes a lot of time. I suggest moving this to section _**after_success**_ to make sure it is run only once.

* The version number in tofu/version.py had to be forced by hand
Indeed, the automated update in _**install**_ section (`export REV=$(python -c "import _updateversion as up; out=up.updateversion(); print(out)"`)) does ot seem to apply at deployment.
I mean that the variable REV is ok, but the version number taken at pipy deployment seems to be not $REV but instead read from tofu/version.py before it was updated by the line above. This suggest that at deployment time, travis re-clones the github repo instead of using the local tofu.

* Apparently the deployment was also done on test-pypi despite the fact I pushed on master and not on test-deploy. My interpretation: I think that a tag in git is not just a commit. When we create a tag, git creates a dedicated branch. The tag is a branch itself. So when travis sees there is a tag, maybe it does not check the branch.

Highest priority: 
------------------
* fix anaconda_upload.sh (deployment today early afternoon if possible)
* Fix multiple `conda build conda_recipe` to save time (should be easy)

Lowest priority:
-----------------
* tofu version update: we can work like this: i.e. by forcing version by hand for 1.4.2, but it would be nice to fix it somehow for the next release 1.4.3 (probably a bit tricky and time-consuming, so no need to do it now)
* Clarify tag vs branch condition in travis (master vs deploy-test)

Tip:
----
@lasofivec , I had to make several tries, if necessary you can work directly on master
If necessary, you can re-deploy the same tga by deleting the existing tag locally, pushing the deletion to the remote, committing changes, and re-tagging and re-pushing tags:

Delete tag locally and push deletion
```
git tag -d 1.4.2
[...]

Issue #346, opened 2020-02-06
tofucalc should handle mat files


Didou09:
For the needs of the data treatment chain at IRFM, tofucalc should be able to handle mat files as input and output (faster data transfer than through IMAS for recursive loops for example)

_**hint:**_ re-use .mat interface from `tfw_METIS_SynthDiags.py`

Didou09:
Fixed in devel by PR #352

PR #348, opened 2020-02-06
Deploy test

lasofivec:
This PR should fix a couple of issues:

- Issue #340 :  _The version number in tofu/version.py had to be forced by hand_
  The problem came from the fact that travis checkouts the tag commit of tofu so the repository is not a specific branch but on a detached HEAD. I changed the `setup.py` so that it checks if it is building a travis tag version (by checking if `$TRAVIS_TAG` exists)
 https://github.com/ToFuProject/tofu/blob/d16595dbca1bc8b516f03abced987c3c2fa2f460/setup.py#L168-L169

- Issue #340 : _Apparently the deployment was also done on test-pypi_
As I mentioned in the issue, the condition `on: tags: True` sets TRAVIS_BRANCH to the tag number, thus it is no longer `master` nor `deploy-test` (nor any other branch name).
Solved this by checking to which git branch the commit hash corresponds to:

https://github.com/ToFuProject/tofu/blob/d16595dbca1bc8b516f03abced987c3c2fa2f460/.travis.yml#L64-L65

and then checking the result as a condition for deployment

https://github.com/ToFuProject/tofu/blob/d16595dbca1bc8b516f03abced987c3c2fa2f460/.travis.yml#L96

There is one caveat for this to work properly, that should still be the proper way to release (will update the documentation):
It is necessary to first push the commit and then push the tag (if you only push the tag it will not find a branch corresponding to that tag so there will be no deployment). Thus, the workflow for a release should be something similar to:

```
# make changes
pip install -e .[dev] # check that it builds
nosetest # check that nothing is broken
git commit -m " some commit message"
git tag -a x.y.z -m "some tag comment"
git push origin
git push --tags
```

- Even if anaconda deployment was not being done, the recipe was still being build. which makes the whole process for nothing, so now it is being done in the upload script

- Issue #292 : the file `_updateversion.py` was no longer being deployed with pip:
https://github.com/ToFuProject/tofu/issues/292#issuecomment-582427149
I used the second solution: I took out out `build-backend = "setuptools.build_meta"` in `pyproject.toml`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,6 @@ install:
 - export IS_MASTER=$(git ls-remote origin | grep "$TRAVIS_COMMIT\s\+refs/heads/master$" | grep -o "master")
 - export IS_DEPLOY=$(git ls-remote origin | grep "$TRAVIS_COMMIT\s\+refs/heads/deploy-test$" | grep -o "deploy-test")
 - echo $TRAVIS_COMMIT
-- git ls-remote origin
 - echo $IS_MASTER
 - echo $IS_DEPLOY
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,7 @@ install:
 - export REV=$(python -c "import _updateversion as up; out=up.updateversion(); print(out)")
 - export VERSION=$(echo $REV | tr - .)
 - echo $REV
+- conda install -q --file requirements.txt
 - python setup.py install
 - export IS_MASTER=$(git ls-remote origin | grep "$TRAVIS_COMMIT\s\+refs/heads/master$" | grep -o "master")
 - export IS_DEPLOY=$(git ls-remote origin | grep "$TRAVIS_COMMIT\s\+refs/heads/deploy-test$" | grep -o "deploy-test")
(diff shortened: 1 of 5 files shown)


lasofivec:
If you agree with these changes, I would add them to release 1.4.2, what do you think ?

lasofivec:
is it ok with you if I try adding this to `master`(v 1.4.2) directly ?

Didou09:
Since we have several PR to merge to master, I suggest we redo a release of 1.4.2 (delete the tag, retag, re-deploy) through a proper PR from devel to master.

I would just like to solve issue #346 before that, hopefully this afternoon, ok for you ?
It's not very clean (deleting tags, retagging...) but we are tuning our release worflow and it shoud get cleaner in the future.

Or, we can start now a cleaner workflow:
* From now on when we do an important release (new major features like the volumes of sight, solid angles...) we increment the second number (1.5.0)
* For all minor features, small improvements, debugging, we increment the third number (1.4.2 => 1.4.3)
* So we could start this afternoon by tagging 1.4.3 instead of deleting 1.4.2

What do you think ?

lasofivec:
Sounds good to me :)

PR #350, opened 2020-02-07
Issue349 new imas reflectometer profile

Didou09:
Main changes:
----------------
* The IMAS data dictionnary has changed a bit since the last version, this is a minor update for ids reflectometer_profile
* Made sure tofuplot and tofucalc remain operational even without installation


tofu/scripts/tofucalc.py
```diff
@@ -13,8 +13,9 @@
 # tofu
 # test if in a tofu git repo
 _HERE = os.path.abspath(os.path.dirname(__file__))
+_HERE = os.path.dirname(os.path.dirname(_HERE))
```
lasofivec: where is _HERE first defined ?

tofu/scripts/tofuplot.py
```diff
@@ -13,8 +13,9 @@
 # tofu
 # test if in a tofu git repo
 _HERE = os.path.abspath(os.path.dirname(__file__))
+_HERE = os.path.dirname(os.path.dirname(_HERE))
```
lasofivec: same question...

tofu/scripts/tofucalc.py
Didou09: one line above

tofu/scripts/tofuplot.py
Didou09: same answer ;-)

tofu/scripts/tofucalc.py
Didou09: Here we are taking _HERE as the directory of the module:
`_HERE = os.path.abspath(os.path.dirname(__file__))`

and then we go 2 parents directories up:
`_HERE = os.path.dirname(os.path.dirname(_HERE))`

That's because tofuplot and tofucalc were moved from the root directory down to tofu/scripts

lasofivec: yes sorry, somehow i didn't see the line above (twice)

Issue #351, opened 2020-02-20
User-customizable imas shortcuts


Didou09:
The default shortcuts used by `tf.imas2tofu.MultiIDSLoader` may vary frequently due to regular updates to the **IMAS** data dictionnnary.
Users should be able to update the **shortcuts** themselves without having to wait for an official to tofu.

Currently, the **shortcuts** are stored in a (large) hard-coded dictionnary in a dedicated module.
An elegant solution would be to mimic ipython's way of customizing preferences:
* create a script `tofu-custom`  that can copy a local version of that module in the user's home, in a dedicated hidden directory: `~/.tofu/`
* copy in this directory the module with an explicit name `_imas2tofu_shortcuts.py`
* Make sure that when tofu imports the dict from the module, it looks first into this local version and then into the general install, to make sure the user's version has precedence over the general version

PR #353, opened 2020-02-21
Issue #351 tofu-custom

Didou09:
Motivations:
-------------
The default shortcuts used by tf.imas2tofu.MultiIDSLoader may vary frequently due to regular updates to the IMAS data dictionnnary.
Users should be able to update the shortcuts themselves without having to wait for an official to tofu.

Currently, the shortcuts are stored in a (large) hard-coded dictionnary in a dedicated module.
An elegant solution would be to mimic ipython's way of customizing preferences:

create a script tofu-custom that can copy a local version of that module in the user's home, in a dedicated hidden directory: `~/.tofu/`
copy in this directory the module with an explicit name _imas2tofu_shortcuts.py or something similar.
Make sure that when tofu imports the dict from the module, it looks first into this local version and then into the general install, to make sure the user's version has precedence over the general version

Main changes:
----------------
* A new script tofu/scripts/tofucustom.py has been created, and an associated bash command `tofu-custom` implemented in the `setup.py`.
* When the user runs it, it automatically creates a hidden local directory `~/.tofu/` in the user's home, and copies files containing default tofu parameters in it (at this point there is only one file `_imas2tofu_def.py`, but more may come in the future). The user is free to edit these files.
* The corresponding tofu sub-package `tofu.imas2tofu` first checks whether the user has local default parameters before loading the general tofu parameters.

N.B.: the names of the files is `_modulename_def.py`

Issues:
-------
Fixes, in devel, Issue #351


setup.py
```diff
@@ -280,7 +280,7 @@ def get_version_tofu(path=_HERE):
     # The project's main homepage.
     url="https://github.com/ToFuProject/tofu",
     # Author details
-    author="Didier VEZINET",
+    author="Didier VEZINET and Laura MENDOZA",
```
lasofivec: :)

tofu/imas2tofu/__init__.py
```diff
@@ -11,8 +11,10 @@
 try:
     try:
         from tofu.imas2tofu._core import *
+        from tofu.imas2tofu._mat2ids2calc import *
     except Exception:
         from ._core import *
+        from ._mat2ids2calc import *
```
lasofivec: is there a reason why you do an "import *" ? is it just because the file is small?
I'm still approving the PR :)

Didou09: It's because the file is small and also because it contains a very specific use-case that may get generalized later, in which case we'll be happy to have it in a separate module that we can import (becasue it would then get bigger)

Thanks !

Issue #347, opened 2020-02-06
_checkformat_inputs_dconfig error in travis


lasofivec:
I have been playing a lot with the travis.yml but branch `deploy-test` is failing with the following error:

> ======================================================================
> FAIL: Failure: AssertionError (Arg config must have at least a StructIn subclass !)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/travis/miniconda/lib/python3.6/site-packages/nose/failure.py", line 39, in runTest
>     raise self.exc_val.with_traceback(self.tb)
>   File "/home/travis/miniconda/lib/python3.6/site-packages/nose/loader.py", line 417, in loadTestsFromName
>     addr.filename, addr.module)
>   File "/home/travis/miniconda/lib/python3.6/site-packages/nose/importer.py", line 47, in importFromPath
>     return self.importFromDir(dir_path, fqname)
>   File "/home/travis/miniconda/lib/python3.6/site-packages/nose/importer.py", line 94, in importFromDir
>     mod = load_module(part_fqname, fh, filename, desc)
>   File "/home/travis/miniconda/lib/python3.6/imp.py", line 235, in load_module
>     return load_source(name, filename, file)
>   File "/home/travis/miniconda/lib/python3.6/imp.py", line 172, in load_source
>     module = _load(spec)
>   File "<frozen importlib._bootstrap>", line 684, in _load
>   File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
>   File "<frozen importlib._bootstrap_external>", line 678, in exec_module
>   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
>   File "/home/travis/miniconda/lib/python3.6/site-packages/tofu-1.4.2b4-py3.6-linux-x86_64.egg/tofu/tests/tests01_geom/tests03_core.py", line 647, in <module>
>     Exp=_Exp, Diag='Test', SavePath=_here)
>   File "/home/travis/miniconda/lib/python3.6/site-packages/tofu-1.4.2b4-py3.6-linux-x86_64.egg/tofu/geom/_core.py", line 3610, in __init__
>     super(Rays, self).__init__(**kwdargs)
>   File "/home/travis/miniconda/lib/python3.6/site-packages/tofu-1.4.2b4-py3.6-linux-x86_64.egg/tofu/utils.py", line 1399, in __init__
>     self._init(**kwdargs)
[...]

PR #355, opened 2020-02-25
Better error message for issue #347

Didou09:
Motivation:
-------------
An error occured (cf. issue #347) on travis for a build, specifying that a config instance was missing a Struct of the subclass StructIn.
The source of this error could not be found, and it does not seem reproducible, as another push of the same branch, with a more explicit error message, did not trigger the error again (all builds successful).
As a result, the associated issue is closed (until it shows up again), and this PR simply aims at implementing a more explicit error message to help debug if this issue pops up again in the future.

Main changes:
----------------
Cleaner code and better error message in the incrimated method

Issues:
-------
Issue #347 is closed until further notice

### DIFF
--- a/tofu/geom/_core.py
+++ b/tofu/geom/_core.py
@@ -3901,20 +3901,31 @@ class Rays(utils.ToFuObject):
 
     @staticmethod
     def _checkformat_inputs_dconfig(config=None):
-        C0 = isinstance(config, Config)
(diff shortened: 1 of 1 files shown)


lasofivec:
@Didou09 did you try to change the installation method again in `setup.py` and see if the error comes back ?

lasofivec:
In any case the PR seems like a good improvement so I am merging

Issue #366, opened 2020-03-11
Half-successful deployment of 1.4.3


Didou09:
Deployment of tofu 1.4.3 was a half-success (see [build](https://travis-ci.org/github/ToFuProject/tofu/builds/661020789))  

[Pypi](https://pypi.org/project/tofu/1.4.3/#files):
------
* sdist to pypi ok
* bdist ok for mac (3.7 and 3.6)
* bdist failed for linux

[Anaconda](https://anaconda.org/ToFuProject/tofu/files):
------------
* anaconda for linux was not attempted because bdist failed for linux
* anaconda for Mac ok

Is there a way to make sure a failed deployment does not stop the other deployments (linux cases) ?

lasofivec:
For `bdist` distribution on Linux machines: as there are so many different linux distributions, with different compilers, etc. for the moment pypa doesn't distribute wheels (already compiled code) for linux machines, but somehow the conditions for the deployment were not met and it still try to deploy it as such. Solving this in a new PR

Didou09:
Fixed, in devel, by PR #371

PR #371, opened 2020-03-30
Travis bdist deployment on master

lasofivec:
I think because of the fact the `conditions` were separated for the `master` branch, travis still tried to deploy with `bdist` for linux machines. To avoid any confusion I put both instructions in one line.
Solves #366


lasofivec:
Awesome, I wont erase the branch as it stays here for future deployment tests

PR #384, opened 2020-05-06
long version name in doc title, fix copyright date future proof

flothesof:
Hi all,

here are some short improvements to the docs:

- fix the version number (right now Tofu 1.4 --> should be 1.4.3)
- fix the copyright notice (© Copyright 2016-2019, Tofu contributors. --> 2016 till current year)
- make use of the HTML last updated thing so that users can see when the HTML page has last been updated.

What do you think?

Regards,
Florian

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -12,6 +12,7 @@
 # serve to show the default.
 import os
 import sys
+import time
 import tofu as tf
 import sphinx_bootstrap_theme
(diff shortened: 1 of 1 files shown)


Didou09:
Merging, I can't get in touch with @lasofivec lately, I think she might be off for some weeks, so assuming I can take the decision alone because this PR has low impact on the core code

Issue #403, opened 2020-06-10
tofu bash command


Didou09:
For conformity with other libraries standards, and ease of use by end-users, it would be nice to make all tofu bash commands accessible through a unique tofu command.

Desired behaviour:
---------------------
* tofu plot **kwdargs => tofuplot **kwdargs
* tofu calc **kwdargs => tofucalc **kwdargs
* tofu custom => tofu-custom
* tofu --version => tofu-version

PR #404, opened 2020-06-11
Issue #403 : Single tofu bash command

Didou09:
Motivation:
-------------
* Since the last release, tofu offers several bash commands:

- `tofu-version`:     to get the current version number
- `tofu-custom`:     to create a local hidden .tofu repo for user-spefific default values
- `tofuplot`:            to load and plot diagnostic data from imas
- `tofucalc`:            to load profiles, diag geometry and equilibrium from imas, calculate synthetic diag data and plot it

* In order to make it easier for users and to mimic standard conventions, it makes sense to implement a single tofu command, and the arguments will determine which of the sub-commands above is called:
- `tofu --version`     for tofu-version
- `tofu custom`        for tofu-custom
- `tofu plot` ....         for tofuplot ...
- `tofu calc`...           for tofucalc ...

Issues:
-------
Fixes, in devel, issue #403

### DIFF
--- a/tofu/entrypoints/tofucalc.py
+++ b/tofu/entrypoints/tofucalc.py
@@ -11,6 +11,12 @@ plt.switch_backend('Qt5Agg')
 plt.ioff()
 
 
+# import parser dict
+sys.path.insert(1, _TOFUPATH)
+from scripts._dparser import _DPARSER
+_ = sys.path.pop(1)
+
+
 # tofu
 # test if in a tofu git repo
 _HERE = os.path.abspath(os.path.dirname(__file__))
@@ -26,12 +32,6 @@ if istofugit:
     _ = sys.path.pop(1)
 else:
     import tofu as tf
-
-
-# import parser dict
-sys.path.insert(1, _TOFUPATH)
-from scripts._dparser import _DPARSER
-_ = sys.path.pop(1)
 
 
 # tforigin = tf.__file__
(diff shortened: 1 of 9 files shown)


Didou09:
Oui, les autres commandes restent dispos aussi, et je suis en train d'en préparer une autre.
Cette PR sera incluse dans la prochaine release

tofu/entrypoints/tofucalc.py
```diff
@@ -3,13 +3,20 @@
...
+
 
 # Generic
 import matplotlib.pyplot as plt
 plt.switch_backend('Qt5Agg')
 plt.ioff()
 
+
+# import parser dict
+sys.path.insert(1, _TOFUPATH)
```
lasofivec, line 15: Is `_TOFUPATH` defined here ? seems like it's defined on line 23

tofu/entrypoints/tofuplot.py
```diff
@@ -3,13 +3,20 @@
...
+
 
 # Generic
 import matplotlib.pyplot as plt
 plt.switch_backend('Qt5Agg')
 plt.ioff()
 
+
+# import parser dict
+sys.path.insert(1, _TOFUPATH)
```
lasofivec, line 15: same question, `_TOFUPATH` seems undefined her

lasofivec (commented):
Okay, but I am not sure there won't be problems with how the variable `_TOFUPATH` is defined in all files

tofu/entrypoints/tofuplot.py
Didou09: You are absolutely right @lasofivec , I had missed this and it was fixed by the following PR #405 👍 

Good to see you here :-) !

Issue #431, opened 2020-12-09
tofu 1.4.8 fails for Windows only on conda-forge


Didou09:
The attempt to export tofu 1.4.8 to conda forge reveals that all unit tests are passed except for [Windows platforms](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=249918&view=logs&j=533b006e-c27f-58a3-8df4-b5a79e785988&t=dccf4699-a398-5507-50c7-f61f6dad5a20) and for python 3.6 and 3.7 (3.8 and 3.9 are fine)

There are 2 errors in the surface and volume sampling routines:
```
======================================================================
ERROR: tests.tests01_geom.tests01_GG.test07_Ves_Vmesh_Tor
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:\bld\tofu_1607509981720\_test_env\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "D:\bld\tofu_1607509981720\test_tmp\tofu\tests\tests01_geom\tests01_GG.py", line 309, in test07_Ves_Vmesh_Tor
    margin=1.e-9)
  File "tofu\geom\_GG.pyx", line 1102, in tofu.geom._GG._Ves_Vmesh_Tor_SubFromD_cython
ValueError: Buffer dtype mismatch, expected 'long' but got 'long long'

======================================================================
ERROR: tests.tests01_geom.tests03_core.Test01_Struct.test15_get_sampleV
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:\bld\tofu_1607509981720\test_tmp\tofu\tests\tests01_geom\tests03_core.py", line 452, in test15_get_sampleV
    algo='new')
  File "D:\bld\tofu_1607509981720\_test_env\lib\site-packages\tofu\geom\_core.py", line 1432, in get_sampleV
    return _comp._Ves_get_sampleV(*args, **kwdargs)
  File "D:\bld\tofu_1607509981720\_test_env\lib\site-packages\tofu\geom\_comp.py", line 712, in _Ves_get_sampleV
    num_threads=num_threads,
  File "tofu\geom\_GG.pyx", line 1102, in tofu.geom._GG._Ves_Vmesh_Tor_SubFromD_cython
ValueError: Buffer dtype mismatch, expected 'long' but got 'long long'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #432, opened 2020-12-09
fix issue 431

flothesof:
It seems that this fixes the issues in #431 on my windows machine.

My explanation:
- under linux np.argmax does indeed return longs
- under windows it doesn't 

Solution:
- explicitly convert dtype


Didou09:
Great ! 
Thanks for the hotfix !
Let's just wait for @lasofivec opinion before merging (she'll have a look this afternoon)

lasofivec (approved):

lasofivec:
Thank you Florian !!

PR #449, opened 2020-12-17
[cleaning up] taking out random notebooks (?)

lasofivec:
I recently noted that github detects tofu like 58% jupyter notebook. I couldn't understand how it's possible, and then I found various notebooks that seem to me that shouldn't be in the library ?


Didou09:
You're right, good point, the notebooks shoudn't be part of the library.

Issue #450, opened 2020-12-18
tofuplot and tofucalc should understand t0 = False


Didou09:
The following command does not work on IRFM Intra, it should:

`tofu plot -i ece  -s 56420 -t0 False`

It's because t0 = False is not properly handled by tofuplot and tofucalc in their respective files

PR #451, opened 2020-12-18
Issue450 t0 false

Didou09:
Main changes:
----------------
* `t0 `is now handled like all other kwdargs by `tofuplot `and `tofucalc `(using `_dparser.py`) and is now a default parameter in imas2tofu (and can thus be customized by the user using `tofu custom`)
* Default `t0 `values for `tofuplot `and `tofucalc `is now `None `to make sure it defaults to imas2tofu default value
* The same is done for `user`, `tokamak`, `run`...
* `Multi._get_t0()` has better error messages and warnings 

Issues:
--------
* Fixes, in devel, issue #450


Didou09 (commented):
All fine for me, @lasofivec  if you can have a look for safety it's good otherwise let me know

tofu/imas2tofu/_core.py
```diff
@@ -1487,19 +1487,22 @@ def get_events(self, occ=None, verb=True, returnas=False):
...
-        if verb:
+        if verb is True or returnas is str:
             msg = np.array([range(times.size), names, times], dtype='U').T
             length = np.nanmax(np.char.str_len(msg))
             msg = np.char.ljust(msg, length)
             msg = ('index'.ljust(length) + ' name'.ljust(length)
                    + '  time ({})'.format(tunits).ljust(length)
                    + '\n' + ' '.join(['-'*length for ii in [0, 1, 2]]) + '\n'
                    + '\n'.join([' '.join(aa) for aa in msg]))
+        if verb is True:
```
lasofivec, line 1498: `if verb` ?

tofu/imas2tofu/_core.py
```diff
@@ -1487,19 +1487,22 @@ def get_events(self, occ=None, verb=True, returnas=False):
         names, times = names[ind], times[ind]
 
         # print and / or return as list / tuple
-        if verb:
+        if verb is True or returnas is str:
```
lasofivec, line 1490: `if verb or returnas is str` ?

lasofivec (approved):

lasofivec:
mini comments, bout ok for PR

tofu/imas2tofu/_core.py
Didou09: verb is a keyword arg used to activate / deactivate verbosity (i.e.: message printing)

tofu/imas2tofu/_core.py
Didou09: returnas is a keyword arg used as a flag to modulate in which form the result should be returned

tofu/imas2tofu/_core.py
lasofivec: Sorry my comment was not clear enough. I meant could you write only `if verb` instead of `if verb is True`

tofu/imas2tofu/_core.py
lasofivec: same here : `if verb is True or returnas is str` => `if verb or returnas is str` ?

tofu/imas2tofu/_core.py
Didou09: Yes an no...
if the user passes, by mistake, something than a bool ( a int or a float for example) if verb will be true (except if the float is 0).
It depends on the behaviour we want, shall we ask the user to pass a bool or shall anything else work too in case he / she makes a mistake ? 
The first case `if verb is True` is cleaner but maybe a bit strict
The second case `if verb `is more flexible but it may happen that the user doesn't understand what he / she is doing

Honestly I'm not sure which one is the best

tofu/imas2tofu/_core.py
Didou09: Same answer as before  ;-)

Issue #452, opened 2020-12-18
Undefined variable in _plot.py


lasofivec:
`ax2` is not defined in the following line

https://github.com/ToFuProject/tofu/blob/beeab0e0c11dd6c4ad7d6c2a149b859704634eff/tofu/_plot.py#L91

Issue #440, opened 2020-12-11
Updating our CI strategy


lasofivec:
So since I have been away it seems it has become more urgent to update our whole CI approach. 
First of all because Travis is really slow lately. I was looking into adding Windows as an OS in the `travis.yml` but for one push I had to wait 2h in the queue... This was also a (minor?) problem for the latest releases. I don't know what you think but I think it is extremely frustrating to say the least.

It seems that many open source projects are migrating away from Travis. One of the first alternatives I have seen is Github's Actions. I want to read a little bit more into it to see if we can and how easy it is. But, since there are many projects doing this it's well documented:
https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/migrating-from-travis-ci-to-github-actions

Which brings me to my second point:
We probably should finally move from `nose` to `pytest`. We talked about it during our meeting last year and in issue #397

I would suggest trying to do this two things during the rest of the month of december. So we start 2021 with a clean slate.

So potentially closing this issue would close issues: #397 #232 #295 #397 ... and I'm probably missing some

lasofivec:
Some resources on Travis slowness and people migrating out of it (thanks @flothesof for the first two)

https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
https://twitter.com/alex_gaynor/status/1336098972864303104
https://github.com/josdejong/mathjs/issues/2024

flothesof:
Adding some reading about Github Actions testing with Python: https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-python

lasofivec:
Yes.
If I understood correctly Travis slowness should be resolved by January, but I still think it's worth while to have a backup plan.
I was looking to other FOSS and many use either: travis, Github's Actions, Azure or a combination of those.

PR #455, opened 2020-12-20
Setting up Github Actions and lightening Travis

lasofivec:
Three main changes to **tofu**'s CI/CD:

- Travis testing:
   On *all* branches two jobs: bionic 3.7 and trusty 3.6
   Only on `devel` and `master`: all jobs are done (bionic 3.7, trusty 3.6, xenial 3.7 and 3.6, osx 3.7 and 3.6)
- Github actions testing:
   Only on `devel` and `master`: [ubuntu-latest (rn: bionic), mac-latest (rn: osx 10.15), windows-latest (rn: )] x [3.6, 3.7, 3.8]
   On all branches: ubuntu latest with python 3.8
- Github actions deployment:
  Trying pypi seems to be working for test pypi. 
  To be able to test it on tofu for sure, we need to add two  add tokens for PyPi on repo's secret ([see this doc](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#saving-credentials-on-github))

Quick fix for #452 
Solves #440


flothesof:
Hi @lasofivec 
Just a quick question before I try it your PR on my side: do you plan on keeping travis builds on top of github actions? 
Isn't it better to just have GH actions run the CI on the three platforms and switch to just that?

lasofivec:
I was thinking for the moment there is no problem.
Plus I haven't implemented all deployments with GA. What I could do is take out the tests from travis since it seems to be working fine on GA and leave (for the moment) the deployments on travis.

lasofivec:
If you rather no have any overlap between travis and GA, I can make a few changes to the GA and make the PR only when travis can be completely suppressed
But in my experience there might be some oversighted issues with the deployment and I think is better at least for one release to have both travis and GA

Didou09:
I agree with @lasofivec , even if it might be better on the long term to have a unique testing / deployment workflow, I also think we should go smoothly with this migration and have the overlap for at least one release to make sure we understand all possible issues and take the time to fix everything. Better be safe and keep the working solution as long as everything is not rock-solid and tested with GA, in real conditions (i.e.: with a real release).

.github/workflows/complete-matrix.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install flake8 pytest coverage
+        pip install -r requirements.txt
+    - name: Lint with flake8
```
Didou09, line 29: Good initiative 👍

.github/workflows/python-package-conda.yml
```diff
@@ -0,0 +1,44 @@
...
+        conda install flake8
+        # stop the build if there are Python syntax errors or undefined names
+        # too many F82 errors, should uncomment the following line
+        # flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        flake8 . --count --select=E9,F63,F7 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: install tofu
+      run: |
+        pip install -e ".[dev]"
```
Didou09, line 41: This file is called conda but the tofu installation is done with pip, I'm not sure I get the point ?

tofu/_plot.py
```diff
@@ -87,8 +87,8 @@ def _plot_shotoverview_init(ns=1, sharet=True, sharey=True, shareRZ=True,
         else:
             laxt[ii] = fig.add_subplot(axarr[ii,:2], sharex=sht, sharey=shy)
             laxc[ii] = fig.add_subplot(axarr[ii,2], sharex=shRZ, sharey=shRZ)
-            if not shareRZ:
-                ax2.set_aspect('equal', adjustable='datalim')
+            # if not shareRZ:
```
Didou09, line 90: Reason for this commenting ? was it the source of a bug ?

Didou09 (commented):
Good, but a few questions / comments below 👍

.github/workflows/python-package-conda.yml
Didou09: pip ? conda ? 
If you want to install from a conda package you first need to create the pack:
 conda build conda_recipe

And then install from that package
conda install -c conda-bld/<name of the created package>

Didou09 (commented):
More comments / questions

tofu/_plot.py
lasofivec: Yes, sorry I will uncomment it as I treated in the PR: #456

lasofivec: (`ax2` is undefined)

.github/workflows/python-package-conda.yml
lasofivec: mmm you are right, I could build the library with conda, however our `conda-recipe` includes the deployment (we have a `conda-upload.sh` file). I decided to build it with `pip` for simplicity. I am going to rename the file and not use `conda`, would you be ok with that solution ? I still have to write a GA file for conda deployment.

lasofivec:
If you both agree, I think it's ready to be merged...?

.github/workflows/python-package-conda.yml
Didou09: OK for that solution

Didou09:
Could you just do the renaming of the file (conda vs pip), for consistency ?
And Happy New Year !!! :-)

lasofivec:
done, happy new year to you too :)

lasofivec:
>     * Github actions deployment:
>       Trying pypi seems to be working for test pypi.
>       To be able to test it on tofu for sure, we need to add two  add tokens for PyPi on repo's secret ([see this doc](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#saving-credentials-on-github))

Regarding the next release, I wrote here the link that I sent you earlier @Didou09 
From the messages I don't feel like we addressed this issue

PR #461, opened 2021-02-12
Update python-package-pip-testing.yml

lasofivec:
There seems to be a big issue with the cython/numpy versions 
https://github.com/scikit-learn-contrib/hdbscan/issues/457#issuecomment-773671043

i'm trying one of the solutions suggested, if it's not working will probably have to force numpy to 1.20


lasofivec:
If it's OK with you @Didou09 I think we can merge

Didou09:
OK, just to be sure:
* you implemented the --no-build-isolation option for the pip install to make sure this error does not occur during CI
* Have you understood the thread you mentionned ? does it say whether this issue is understood and will be fixed somehow ?

Didou09:
But ok for merging anyway, my questions were more out of curiosity

Issue #459, opened 2021-01-15
Inverted camera ?


Didou09:
In [23]: conf = tf.geom.utils.create_config('TCV')

In [24]: cam = tf.geom.utils.create_CamLOS2D(P=[2.5, 0, 0], F=0.21944, angs=[np.pi, 0, 0], N12=[195, 487], D12=[0.03354, 0.083764], config=conf, Name='SXR', Exp='TCV', Diag='ME-SXR')
In [30]: cam.config.set_colors_random()

In [31]: cam.plot_touch()
Out[31]: <tofu.utils.KeyHandler_mpl at 0x7f6a87ed4e50>

In [32]: cam.plot_touch(invert=False)
Out[32]: <tofu.utils.KeyHandler_mpl at 0x7f6a973ece90>

![image](https://user-images.githubusercontent.com/5929562/104743147-a659d180-574b-11eb-997b-36b915c719e6.png)

Issue #438, opened 2020-12-10
Update WEST config


Didou09:
The WEST config in tofu is detailed, but it could be more accurate, in particular regarding the vessel and inner thermal protection.

PR #462, opened 2021-02-15
Issue #438 west details

Didou09:
More details are added to the WEST geometry, from an excel file provided by Ph. Moreau.
It includes in particular:
* The inner and outer vacuum vessels
* All parts of the divertor coils casings (upper and lower)
* The HFS thermal shield
* The LFS thermal shield divided into 4 non-axisymmetric parts to make sure port plugs are implemented

All new structures are plotted in blue below

image

image

On the second image, it can be noticed that the baffle appears as red while it should appear as grey:
* This could be related to issue #459 ?
* It has to do with wrong structure indexing (see cam.dgeom['indout'] vs cam.dgeom['indStruct'])
=> Opening a dedicated issue


tofu/geom/inputs/WEST_get_details_from_excel.py
```diff
@@ -0,0 +1,347 @@
...
+                           _PPLUG_PTS[:, 1]])
+    _, indspp = np.unique(poly_pplug[1, :], return_index=True)
+    poly_pplug = poly_pplug[:, indspp]
+    poly_pplug[0, :] = np.interp(poly_pplug[1, :], poly[1, :], poly[0, :])
+    ind_insert = np.searchsorted(poly[1, :], poly_pplug[1, :])
+
+    # TBF
+    poly_full = np.insert(poly, ind_insert, poly_pplug, axis=1)
+    pos_full_slim =  _PPLUG_POS[0] + 0.5*_DTHETA + None
+    extent_full =
```
lasofivec, line 206: This is the line that's making all jobs fail

lasofivec (changes requested):

tofu/geom/inputs/WEST_get_details_from_excel.py
Didou09: Yes, you're right, I realised too late there was something wrong.
I'm taking the opportunity to implement more details

Didou09:
All seems fine now @lasofivec , if your review is positive don't hesitate to merge (and delete the branch)

Issue #471, opened 2021-02-16
Wrong input argument definition in _GG.LOS_areVis_PtsFromPts_VesStruct()


Didou09:
As described in PR #470 , and as can be seen from the code in tofu/geom/_comp_solidangles.calc_solidangle_particle(), input argument k (=len_v in calc_solidangle_particle()) is necessarily a 2d array because it contains the distance between all points from 2 sets of (N,) and (M,) points respectively, so it is necessarily a (N, M) array.

But the routine in _GG expects a 1d array.

Also, I can't remember why the routine returns an array of int and not bool ?

lasofivec:
Answering the last question first : the bool type is kind of complicated to translate from python to numpy to cython (use of int, int8, np.bool,c bool, bint, npy_bool...). They don't use the same and if I try to use bool there a lot of transformations needed which slows down the cython part.

Didou09:
> Answering the last question first : the bool type is kind of complicated to translate from python to numpy to cython (use of int, int8, np.bool,c bool, bint, npy_bool...). They don't use the same and if I try to use bool there a lot of transformations needed which slows down the cython part.

OK for me

PR #472, opened 2021-02-17
[#471] fixed array shape of parameter k in are_vis_ptsfrompts

lasofivec:
Bugfix:
`LOS_areVis_PtsFromPts_VesStruct` took the parameter `k` (distance between the first and second sets of points) as a vector instead of a matrix. Now it's fixed.

Solves #471


tofu/geom/_raytracing_tools.pyx
```diff
@@ -2078,7 +2078,7 @@ cdef inline void are_visible_vec_vec(double[:, ::1] pts1, int npts1,
         is_visible_pt_vec(pts1[0,ii], pts1[1,ii], pts1[2,ii],
                           pts2, npts2,
                           ves_poly, ves_norm,
-                          ind[ii,...], k, ves_lims,
+                          ind[ii,...], k[ii,...], ves_lims,
```
Didou09, line 2081: ok, 
Just question before validating:

1. Did you double-check to make sure you're incrementing the good index ? 
To be consistent `k` has to be shaped like `ind`, right ?
Can you just add a docstr to the top-most function (i.e.:  `_GG.LOS_areVis_PtsFromPts_VesStruct()`, the first one visible by the user) to explain how `k` should be shaped ? ( `(pts1.shape[1], pts2.shape[1])` or `(pts2.shape[1], pts1.shape[1])` ?)

2. Did you check the same error was not existing elsewhere ? 
I guess that in `_GG.LOS_isVis_PtFromPts_VesStruct()` `k` is fine as a 1d array, can you confirm ?

Didou09 (commented):
OK, just 2 questions below

tofu/geom/_raytracing_tools.pyx
lasofivec: Sorry I jumped the gun with the PR, I'm going to add a short test in unit test and already updated the doc

tofu/geom/_GG.pyx
```diff
@@ -1099,7 +1099,7 @@ def _Ves_Vmesh_Tor_SubFromD_cython(double rstep, double zstep, double phistep,
     _st.vmesh_prepare_tab(lnp, sz_r, sz_z, sz_phi)
     indI = np.sort(indI, axis=1)
     indi_mv = indI
-    first_ind_mv = np.argmax(indI > -1, axis=1).astype(np.long)
+    first_ind_mv = np.argmax(indI > -1, axis=1).astype(int)
```
Didou09, line 1102: Is this the new tuype introduced by more recent numpy / cython versions ?

tofu/utils.py
```diff
@@ -149,7 +149,7 @@ def flatten_dict(d, parent_key='', sep=None, deep='ref',
                 elif deep=='copy':
                     v = v.copy(deep='copy')
             new_key = parent_key + sep + k if parent_key else k
-            if isinstance(v, collections.MutableMapping):
+            if isinstance(v, collections.abc.MutableMapping):
```
Didou09, line 152: Why going for the abstract class?

Didou09 (commented):
Good job!
Just 2-3 comments / questions below

tofu/utils.py
lasofivec: I got this warning, saw it was possible to use the abstract class without repercussions and ran tests, seems it does work:

`DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working`

tofu/geom/_GG.pyx
lasofivec: yes, it seems to be working on windows as well (Github Actions tests passing)

tofu/utils.py
Didou09: Ok, good job :-)
Thanks

Didou09:
I'm assumaing we can close the branch ssince it was quite specific for this issue, you can always restore it f necessary

Issue #475, opened 2021-03-01
Hunt warnings from unit tests


Didou09:
Units tests are passing, but show several warnings that seem to be relatively simple to remove (minor syntax updates)
See for example [here](https://github.com/ToFuProject/tofu/runs/2004133116?check_suite_focus=true).

PR #476, opened 2021-03-01
Issue #475 Remove some easy warnings

Didou09:
Motivations:
--------------
* The unit tests in [GA](https://github.com/ToFuProject/tofu/runs/2004133116?check_suite_focus=true) showed a few warnings that seemed easy to remove

Main changes:
----------------
* A few minor changes (in particular several occurences of `np.int`  or `np.long` replaced by `int` as recommended [here](https://numpy.org/devdocs/release/1.20.0-notes.html)).

Issues:
-------
* Fixes, in devel, issue #475

### DIFF
--- a/tofu/tests/tests01_geom/test_01_GG.py
+++ b/tofu/tests/tests01_geom/test_01_GG.py
@@ -1797,16 +1797,16 @@ def test24_is_visible(debug=2):
             shot=0
         )
         s1 = tfg.PFC(Name="S1",
-                     Poly=[SP0x, SP0y],
+                     Poly=[SP0x[:-1], SP0y[:-1]],
                      Exp="Misc",
                      shot=0)
         s2 = tfg.PFC(Name="S2",
-                     Poly=[SP1x, SP1y],
+                     Poly=[SP1x[:-1], SP1y[:-1]],
                      Exp="Misc",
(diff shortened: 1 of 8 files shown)


tofu/tests/tests01_geom/test_01_GG.py
```diff
@@ -1797,16 +1797,16 @@ def test24_is_visible(debug=2):
             shot=0
         )
         s1 = tfg.PFC(Name="S1",
-                     Poly=[SP0x, SP0y],
+                     Poly=[SP0x[:-1], SP0y[:-1]],
```
lasofivec, line 1800: why these changes ? is it to leave the polygon open ?

lasofivec (approved):
Ok

tofu/tests/tests01_geom/test_01_GG.py
Didou09: Yes, because the PFC class checks for doubles in the polygons.
If it provided as a closed polygon, it will work, but the class will throw a warning to the user just saying that it detected the double point.
It is a relatively normal behavior (double-checking inputs and keeping the user informed), but for unit tests we already know it and we just want to get rid of easy warnings so really problematic ones are more visible.

Issue #480, opened 2021-03-03
Add dependency requests


Didou09:
Package `requests` is an implicit dependency that we forgot to include, it is necessary for sub-package `openadas2tofu`.

PR #482, opened 2021-03-03
Issue #480 Added requests to setup.py, requiremens.txt and conda_recipe/meta.yaml

Didou09:
Motivations:
--------------
* I realised one dependency was missing: the library `requests`
* It necessary only for the subpackage `openadas2tofu`, which is rarley used and not tested yet, which explains why we did notice it before

Main changes:
----------------
* Added to `setup.py`
* Added to `requirements.txt`
* Added to the run-time section of `conda_recipe/meta.yaml`

I think that's it, except if it is also needed in `pyproject.toml`?
@lasofivec, can you double-check whether all dependencies need to be in `pyproject.toml`? 
Currently `scipy`, `matplotlib` and `requests` are not in there.

Issues:
-------
* Fixes in devel issue #480

### DIFF
--- a/conda_recipe/meta.yaml
+++ b/conda_recipe/meta.yaml
@@ -38,6 +38,7 @@ requirements:
     - Cython >=0.26
     - pytest
     - pygments
+    - requests
 
 test:
   requires:
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,7 @@
 numpy
 scipy
 matplotlib
+requests
 
 ######## Requirements with Version Specifier ########
 Cython>=0.26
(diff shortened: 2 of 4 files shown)


lasofivec:
No, it's fine as it is.
This file installs everything to build the wheel before running the `setup` file. So it only includes what's needed for the setup.py to work.

>  **Now setuptools has a setup_requires argument to specify what is necessary to build a project, but your can't read that setting unless you have setuptools installed, which meant you couldn't declare you needed setuptools to read the setting in setuptools.** This chicken-and-egg problem is why tools like virtualenv install setuptools by default and why pip always injects setuptools and wheel when running a setup.py file regardless of whether you explicitly installed  it. Oh, and don't even try to rely on a specific version of setuptools for buildling your project as there was no way to specify that; you had to make do with whatever the user happened to have installed.

source: https://snarky.ca/what-the-heck-is-pyproject-toml/

lasofivec (approved):

Issue #485, opened 2021-03-09
Handle open polygons from IMAS wall


Didou09:
Some tokamak geometries are stored in very simplified version in IMAS, only as a set of open polygons

Example:
----------

./tofu/entrypoints/tofuplot.py -i wall -tok ITER_MD -u public -r 17 -s 1180

![image](https://user-images.githubusercontent.com/5929562/110474466-fd42ae00-80df-11eb-8cb8-793deff31e9f.png)

tofu should find a way to handle this and avoid closing them automatically when not relevant (introduce a flag ?)

Issue #442, opened 2020-12-15
tofu plot should read and plot ids wallo on ITER


Didou09:
The following subpackages are not available:
    - tofu.mag
  => see print(tofu.dsub[<subpackage>]) for details.
  warnings.warn(msg)
usage: tofu [-h] [-c CONFIG] [-tse3x TOSOLEDGE3X] [-s SHOT [SHOT ...]]
            [-u USER] [-tok TOKAMAK] [-r RUN] [-v VERSION] -i
            {barometry,bolometer,bremsstrahlung_visible,calorimetry,ece,ic_antennas,interferometer,lh_antennas,magnetics,neutron_diagnostic,polarimeter,reflectometer_profile,soft_x_rays,spectrometer_visible,equilibrium,core_profiles,core_sources,edge_profiles,edge_sources,magfieldlines,events,shortcuts,config}
            [{barometry,bolometer,bremsstrahlung_visible,calorimetry,ece,ic_antennas,interferometer,lh_antennas,magnetics,neutron_diagnostic,polarimeter,reflectometer_profile,soft_x_rays,spectrometer_visible,equilibrium,core_profiles,core_sources,edge_profiles,edge_sources,magfieldlines,events,shortcuts,config} ...]
            [-q QUANTITY [QUANTITY ...]] [-X X [X ...]] [-t0 T0] [-t T]
            [-tl TLIM [TLIM ...]] [-ich INDCH [INDCH ...]]
            [-ichauto INDCH_AUTO] [-e EXTRA] [-sx [SHAREX]]
            [-bck [BACKGROUND]] [-mag_sep_dR MAG_SEP_DR]
            [-mag_sep_nbpts MAG_SEP_NBPTS]
            [-mag_init_pts MAG_INIT_PTS MAG_INIT_PTS MAG_INIT_PTS]
tofu: error: argument -i/--ids: invalid choice: 'wall' (choose from 'barometry', 'bolometer', 'bremsstrahlung_visible', 'calorimetry', 'ece', 'ic_antennas', 'interferometer', 'lh_antennas', 'magnetics', 'neutron_diagnostic', 'polarimeter', 'reflectometer_profile', 'soft_x_rays', 'spectrometer_visible', 'equilibrium', 'core_profiles', 'core_sources', 'edge_profiles', 'edge_sources', 'magfieldlines', 'events', 'shortcuts', 'config')

PR #486, opened 2021-03-09
Issue #442 read ids wall iter

Didou09:
Motivation:
------------
* tofuplot was not able to load ids wall 
* The names of the structures stored in ids wall are non-conform (include blank spaces and underscores) and are not handled correctly by tofu

Main change:
---------------
* tofuplot can now handlle ids wall
* All names are now scanned for ' ' and '_' , which are removed and the various parts of the names are capitalized

Example (on hpc-login02):
-----------------------------
./tofu/entrypoints/tofuplot.py -i wall -tok ITER_MD -u public -r 17 -s 1180

image

Issues:
-------
* Issue #442 is fixed in devel
* Issue #485 is opened to address the fact that the polygons stored in IMAS are not closed and represent 2 parts of the same polygon.

### DIFF
--- a/scripts/_dparser.py
+++ b/scripts/_dparser.py
@@ -201,9 +201,10 @@ def parser_plot():
 
     tf, MultiIDSLoader, _defscripts = get_mods()
 
+    _LIDS_CONFIG = MultiIDSLoader._lidsconfig
     _LIDS_DIAG = MultiIDSLoader._lidsdiag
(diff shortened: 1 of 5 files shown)


lasofivec:
I didn't understand how this addresses #485

Didou09:
This doesn't address #485, it is the other way around: this PR revealed issue #485 , which will be addressed later in  adedicated PR

Issue #407, opened 2020-09-07
Struct from various file formats


Didou09:
Make it easy to save to / load from not only txt files but also:
* svg (to draw from inkscae or draw.io)
* csv

PR #511, opened 2021-06-25
Issue407 struct from files

Didou09:
Motivation:
--------------
* Create config from Inkscape, load it to tofu via svg file

Main changes:
--------------------
* `Config.from_svg()` implemented
* User defined in Inkscape
    - class of each struct (no fill => `Ves`, fill => `PFC`)
    - Name of each `Struct` (id)
    - Polygon using Bezier Curves and straight lines tool, they must be closed:
            image

* Adds a new dependence: [svg.path](https://github.com/regebro/svg.path) for svg path parsing

Issue:
---------
* Fixes, in devel, issue #407 

Example:
------------
```
In [1]: import tofu as tf
/home/didier/Documents/Projects/tofu/tofu/__init__.py:92: UserWarning: 
The following subpackages are not available:
	- tofu.imas2tofu
	- tofu.mag
  => see print(tofu.dsub[<subpackage>]) for details.
  warnings.warn(msg)

In [2]: conf = tf.geom.Config.from_svg('tofu/tests/tests01_geom/test_03_core_data/Inkscape.svg', Name='test', Exp='test', res=10)
The following structures were loaded:
	- Ves: vessel (65 pts, None)
	- PFC: limiter (5 pts, #0000ff)
	- PFC: pfcoil (5 pts, #ff0000)
from /home/didier/Documents/Projects/tofu/tofu/tests/tests01_geom/test_03_core_data/Inkscape.svg
/home/didier/Documents/Projects/tofu/tofu/geom/_core.py:481: UserWarning: Struct instance: double identical points in Poly
  => 1 points removed
  => Poly goes from 65 to 64 points
  warnings.warn(msg)
/home/didier/Documents/Projects/tofu/tofu/geom/_core.py:481: UserWarning: PFC instance: double identical points in Poly
  => 1 points removed
  => Poly goes from 5 to 4 points
  warnings.warn(msg)

In [3]: conf
Out[3]: 
tot. Struct  tot. occur  tot. points
-----------  ----------  -----------
3            3           72         

class  Name     SaveName                                                         nP  noccur  move  color                     visible
-----  -------  ---------------------------------------------------------------  --  ------  ----  ------------------------  -------
Ves    vessel   TFG_Ves_Exptest_vessel_sh00000_Vers1.4.10-alpha9-280-g3fdd0810   64  0       None  ( 0.0,  0.0,  0.0,  1.0)  True   
PFC    limiter  TFG_PFC_Exptest_limiter_sh00000_Vers1.4.10-alpha9-280-g3fdd0810  4   0       None  ( 0.0,  0.0,  1.0,  1.0)  True   
PFC    pfcoil   TFG_PFC_Exptest_pfcoil_sh00000_Vers1.4.10-alpha9-280-g3fdd0810   4   0       None  ( 1.0,  0.0,  0.0,  1.0)  True   

```


lasofivec:
> Adds a new dependence: svg.path for svg path parsing

so it should probably be added to the `setup.py` right?

tofu/geom/_core.py
```diff
@@ -1916,50 +1936,73 @@ def from_txt(
...
+        # Check inputs
+        if returnas not in [object, 'object', dict, 'dict']:
+            msg = ("Arg returnas must be either:"
+                   + "\t- 'object': return {} instance\n".format(cls.__name__)
+                   + "\t- 'dict' : return a dict with polygon, pos and extent")
             raise Exception(msg)
-        if not pfe[-4:] == ".txt":
-            msg = "Only accepts .txt files (fed to np.loadtxt) !\n"
-            msg += "    file:  %s" % pfe
+        if pfe[-4:] not in ['.txt', '.csv']:
```
lasofivec, line 1961: I read recentley that a better way to check for file extensions is to do 
``pfe.lower().endswith(('.txt', '.csv'))``
maybe the lower part is unnecessary. But it would alow to check something like:
``pfe.lower().endswith(('.txt', '.csv', '.jpeg', '.m'))``
so no need to cound the number of characters and is more compact and versatile

Didou09: Very good advice, thanks, I'll correct!

Didou09: Good to hear from you !  :-)

Didou09:
> so it should probably be added to the `setup.py` right?

Yes and no....
It is a dependency that is needed only for this particular method, and I don't want the installation to fail just because the user does not have access to this particular dependency, 
But I am not sure we define **_optional_** dependencies.

As a temporary solution I implemented an explicit error message that explains to the user this particular method needs this dependency and he should install it to use the method.

What do you think ?

tofu/geom/_comp.py
```diff
@@ -43,6 +51,145 @@
...
+        raise Exception(msg)
+
+    # Derive usable data
+    kstr = 'fill:'
+    lk = list(dpath.keys())
+    for ii, k0 in enumerate(lk):
+
+        v0 = dpath[k0]
+        poly = _get_pts_from_path_svg(v0['poly'], res=res)
+        # reverse because for some reason the parser inverses y
```
lasofivec, line 160: where is the inversion done? is it necessary?

lasofivec:
First of all, this is amazing! Congrats.

I tried it rn but it looks like it is reversed
Screenshot from 2021-06-25 13-40-16
Screenshot from 2021-06-25 13-40-12

I am using Inkscape 1.1 (exact number: Inkscape 1.1 (ce6663b3b7, 2021-05-25)). I read in the code that you wrote:

>         # reverse because for some reason the parser inverses y

but I don't think you actually reverse `y` or maybe you do and shouldn't?

I will add the library to the setup and would suggest adding a `how to` to the doc?

Didou09:
Thanks !

No, I don't reverse it because I don't know yet why y gets reversed
I opened a [ticket](https://github.com/regebro/svg.path/issues/76#issuecomment-868339729) on the repo of the svg parser used to ask the author if he knows what this is due to, I'm waiting for his answer before choosing the adequate solution.

OK for the howto, excellent idea,
For adding the library to the `setup.py` (and maybe to the `requirements.txt` and to the `conda_recipe/meta.yaml` and the `pyproject.toml`?) ok too, but make sure it doesn't block the deployment in case of issue with this dependency (I had an issue during my firs

lasofivec:
I would suggest to add it to the setup as a `dev` dependency in the `extras_require` so it will only be installed if you install tofu as a `dev`. And then it depends if you are going to add a unit test? If so, I would probably have to add it to the conda recipe as well.

Didou09:
> I would suggest to add it to the setup as a `dev` dependency in the `extras_require` so it will only be installed if you install tofu as a `dev`. And then it depends if you are going to add a unit test? If so, I would probably have to add it to the conda recipe as well.

Yes, goog idea to include it for dev only as a first step.
I've started implementing a unit test, but it is deactivated until I understand why the y coordinates get reversed and until we understand how to make sure it does not break deployment.

If you can take care of the dependency issue that would be great, yiou can work directly on this branch since I will touch it only for the y reversal pb. So if you could:
* Include it as dependency for dev only first
* run tests in GA and Travis
* Try adding it as a full dependency 
* run tests in GA and travis and see if all goes welll (there might be a pb here)
Thanks !

lasofivec:
I created a PR #512 with it as a `dev` requirement, no need to do it as a main dependency since I think as you say it is a small feature not needed for everybody nor all the time. From what I have read online this is the way to go. I created a PR from my forked version so the GA and travis will be tested.

lasofivec:
Looks like you got an answer (https://github.com/regebro/svg.path/issues/76#issuecomment-868468396). If you flip the y axis and merge my PR, I would add tuto in the doc (maybe a small video will be nice?), this would include:

- creation of vessel
- creation of structures
- changing path properties to set names
- creating, loading and plotting configurations with tofu

would you like to add something?

Didou09:
Correction done,
I've also added an optional `z0` argument so the user can choose where to place the z origin:

```
In [2]: conf = tf.geom.Config.from_svg('tofu/tests/tests01_geom/test_03_core_data/Inkscape.svg', Name='test', Exp='test', res=10, z0=-150)
The following structures were loaded:
	- Ves: vessel (65 pts, None)
	- PFC: limiter (5 pts, #0000ff)
	- PFC: pfcoil (5 pts, #ff0000)
from /home/didier/Documents/Projects/tofu/tofu/tests/tests01_geom/test_03_core_data/Inkscape.svg
/home/didier/Documents/Projects/tofu/tofu/geom/_core.py:481: UserWarning: Struct instance: double identical points in Poly
  => 1 points removed
  => Poly goes from 65 to 64 points
  warnings.warn(msg)
/home/didier/Documents/Projects/tofu/tofu/geom/_core.py:481: UserWarning: PFC instance: double identical points in Poly
  => 1 points removed
  => Poly goes from 5 to 4 points
  warnings.warn(msg)

In [3]: conf.plot()

```

image

Didou09:
> Looks like you got an answer ([regebro/svg.path#76 (comment)](https://github.com/regebro/svg.path/issues/76#issuecomment-868468396)). If you flip the y axis and merge my PR, I would add tuto in the doc (maybe a small video will be nice?), this would include:
> 
>     * creation of vessel
> 
>     * creation of structures
> 
>     * changing path properties to set names
> 
>     * creating, loading and plotting configurations with tofu
> 
> 
> would you like to add something?

That would be perfect thanks !
Now that everything seems to be set, can I let you deal with the merging and both PR and with the doc? I have a meeting started in 30 s :-)

Issue #513, opened 2021-06-25
Error when loading AUG config


lasofivec:
So I was trying to update the doc when I noticed the following script was failing `examples/tutorials/tuto_plot_gallery_fusion_machines.py` (also in tests). First, there are many 

> /home/lasofivec/tofu/tofu/geom/utils.py:854: UserWarning: Could not be loaded: TFG_PFC_ExpAUG_D2cdome.txt

and it ends with

> 
> Traceback (most recent call last):
>   File "examples/tutorials/tuto_plot_gallery_fusion_machines.py", line 40, in <module>
>     config.plot()
>   File "/home/lasofivec/tofu/tofu/geom/_core.py", line 3777, in plot
>     vis = self.get_visible()
>   File "/home/lasofivec/tofu/tofu/geom/_core.py", line 2891, in <lambda>
>     setattr(self, "get_%s" % pp, lambda pk=pp: self._get_extraprop(pk))
>   File "/home/lasofivec/tofu/tofu/geom/_core.py", line 2806, in _get_extraprop
>     k0, k1 = self._dStruct["lorder"][0].split('_')
> IndexError: list index out of range
> 

The actual error is because no file was read successfully.
 From what my investigation is coming from the following lines

https://github.com/ToFuProject/tofu/blob/f0697ffb2c9afa93ab31f4be31c0d97496303f78/tofu/geom/_core.py#L1968-L1977

AUG is the only configuration for which the data is separated by several white spaces, and as the delimiter is marked as one white space and not as None which NumPy interprets as one or several whitespaces, then the second column is thought to be a white-space and the actual data is ignored.

## Remark
What I haven't understood yet is how this was not caught before nor in the CI. But I will look into it on Monday.
Might be a new version of numpy? 

## Versions
numpy version: 1.20.1
tofu branch: Issue407_StructFromFiles
tofu version: 1.4.10-alpha9-280-g3fdd0810
(I tried with devel as well, same error)

## A possible solution:
https://github.com/ToFuProject/tofu/blob/f0697ffb2c9afa93ab31f4be31c0d97496303f78/tofu/geom/_core.py#L1972
just use `None`

Didou09:
So if we use None it solves the problem?

PR #516, opened 2021-06-29
Issue513 aug config not loading

Didou09:
Motivations:
--------------
*  As reported by @lasofivec , the AUG config was not loading properly in the tutorials
* It was due to a too constraining `delimiter `in `numpy.loadtxt()` AUG files had 2 white spaces as separators)

Main changes:
----------------
* The delimiter is set to `None`, which solves the issue
* But since this error was not caught by unit tests previously, a new dedicated unit test has been implemented to catch it if it shows up again
* It relies on the introduction of a new keyword argument `tf.load_config('AUG', strict=True)` that forces forces an `Exception `(instead of a simple `warning`) to be raised if a `Structure` cannot be loaded properly
* The associated error message was refactored to be more informative and more compact (for readability)
* Thanks to the new unit test, another hidden bug was identified: an old useless config was still considered as valid except that the corresponding files had been deleted. This config is now removed.

Issues:
--------
* Fixes, in devel, issue #513 

Example:
----------

The following works fine now
```
In [1]: import tofu as tf                                                                                                                                                              
/Home/DV226270/ToFu_All/tofu_git/tofu/tofu/imas2tofu/__init__.py:87: UserWarning: 
You do not seem to be using the latest IMAS version:
'module list' vs 'module av IMAS' suggests:
	- Current version: 3.30.0-4.8.4
	- Latest version : 3.32.1-4.9.1-R2019b
  warnings.warn(msg)

In [2]: conf = tf.load_config('AUG')   
```

### DIFF
--- a/tofu/geom/_core.py
+++ b/tofu/geom/_core.py
@@ -1969,7 +1969,7 @@ class Struct(utils.ToFuObject):
             if pfe[-4:] == '.csv':
                 delimiter = ', '
             else:
-                delimiter = ' '
+                delimiter = None
         if comments is None:
             comments = _COMMENT
 
--- a/tofu/geom/_def_config.py
+++ b/tofu/geom/_def_config.py
@@ -169,11 +169,11 @@ _DCONFIG = {
         'CoilPF': ['PF1', 'PF2', 'PF3', 'PF4', 'PF5', 'PF6'],
         'CoilCS': ['CS3U', 'CS2U', 'CS1U', 'CS1L', 'CS2L', 'CS3L'],
     },
-    'ITER-SOLEDGE3XV0': {
-        'Exp': _ExpITER,
-        'Ves': ['SOLEDGE3XV0'],
-        'PFC': ['SOLEDGE3XDivDomeV0', 'SOLEDGE3XDivSupportV0']
-    },
+    # 'ITER-SOLEDGE3XV0': {
+        # 'Exp': _ExpITER,
+        # 'Ves': ['SOLEDGE3XV0'],
+        # 'PFC': ['SOLEDGE3XDivDomeV0', 'SOLEDGE3XDivSupportV0']
+    # },
 
     # AUG
     'AUG-V1': {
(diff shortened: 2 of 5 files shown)


tofu/geom/_def_config.py
```diff
@@ -169,11 +169,11 @@
...
-    'ITER-SOLEDGE3XV0': {
-        'Exp': _ExpITER,
-        'Ves': ['SOLEDGE3XV0'],
-        'PFC': ['SOLEDGE3XDivDomeV0', 'SOLEDGE3XDivSupportV0']
-    },
+    # 'ITER-SOLEDGE3XV0': {
+    # 'Exp': _ExpITER,
+    # 'Ves': ['SOLEDGE3XV0'],
+    # 'PFC': ['SOLEDGE3XDivDomeV0', 'SOLEDGE3XDivSupportV0']
+    # },
```
lasofivec, line 176: why not just erase these lines?

tofu/geom/utils.py
```diff
@@ -819,51 +833,83 @@ def _create_config_testcase(
...
-                  divlow=None, divup=None, nP=None,
-                  returnas=None, SavePath='./', path=_path_testcases):
+def create_config(
+    case=None, Exp='Dummy', Type='Tor',
+    Lim=None, Bump_posextent=[np.pi/4., np.pi/4],
+    R=None, r=None, elong=None, Dshape=None,
+    divlow=None, divup=None, nP=None,
+    returnas=None, strict=None,
+    SavePath='./', path=_path_testcases,
+):
```
lasofivec, line 912: Could you update the doc below? since this function is directly used by the users

tofu/geom/_def_config.py
Didou09: In case we need to create it again in the future, it's just a precaution.

tofu/geom/utils.py
Didou09: Done   :-)

tofu/geom/_core.py
```diff
@@ -1969,7 +1969,7 @@ def from_txt(
             if pfe[-4:] == '.csv':
```
lasofivec, line 1969: and just because we are already changing here, would you mind putting `pfe.endswith(".csv")`? (If you liked that solution)

Didou09: Done, I agree this is cleaner

tofu/geom/utils.py
Didou09: P.S.: @lasofivec  I would like to release 1.4.11 today or tomorrow, shall I wait for a PR on your side?

lasofivec: yes, coming in a couple of minutes

lasofivec (approved):

tofu/geom/utils.py
Didou09: OK, I'll do it at the end of the day, probably around 16h-17h, so take your time to polish the PR :-)

PR #517, opened 2021-06-29
Added tutorial "how to create geometry from svg"

lasofivec:
Created a video and a tutorial in the website to show how to create a geom using inkscape. The video is uploaded to my youtube channel, I don't know if `tofu` has a google account? or a youtube channel? but if so we can put it there

https://www.youtube.com/watch?v=BoZh2U7I8-g

I tried to link it from a local file but it was a bit difficult without using a sphinx extension.

Updated today with latest changes (possibility to change `scale` and `r0` and now it's failing. See

https://tofuproject.github.io/tofu/auto_examples/tutorials/tuto_plot_create_geometry_from_svg.html#sphx-glr-auto-examples-tutorials-tuto-plot-create-geometry-from-svg-py

just loading the file works:

Screenshot from 2021-06-29 14-01-58

but uploading with r0 and scale:

Screenshot from 2021-06-29 14-02-12


Didou09:
The video is perfect!
Thanks :-)   👍 

tofu does not have a youtube channel, although it probably should at some point.
In the meantime you we can include a link to your youtube channel in the doc ?

I'll look at the bug

Didou09:
I see you managed to embed the video in the doc from your channel, that's a very good solution  👍

Didou09:
Hum... I think the bug is because you choose a r0 such that the 2 structures vessel and struct1 have negative r values, which is not possible in toroidal geometry because R is always positive.

I'll add a safety check and explicit warning in the code, in a new PR, 

In the meantime can you retry with r0 = 20 and r0 = -100 to check ?
r0 can basically be used to change the aspect ratio of the tokamak.

Didou09:
Question: what tools do you use to make that kind of tutorial video?

We should probably do more of them 👍

lasofivec:
Oh, you are right, I was trying to center the poloidal cut and thus creating a badly defined torus! 
Thanks!

lasofivec:
For the video, I struggled a little bit. Mostly because I wanted to have a version with the subtitles on the video (not as additional track). But what I did was
- screen record with [Simple Screen Recorder](https://www.maartenbaert.be/simplescreenrecorder/) (Recording was a breeze but doing all the steps without errors was time consuming 😅 )
- I uploaded first to youtube and used their tool to write CC captions for instructions
- Downloaded the subtitles
- Converted the video from mkv to mp4
- Used this answer https://askubuntu.com/a/1273657/226616 to "merge" the subtitles to the video

Didou09:
Idea:
------
We could use this new svg functionality to show users how they copy copy a published tokamak geometry and easily reproduce it in tofu:
* open inkscape and import an image of a tokmaka poloidal cross-section taken from a published article
* draw the elements you are interested it 
* save as svg
* open ipython, import tofu, load config from svg

What do you think ?
Do you think you could make a tuto video using fig 1 of the article attached (DIIID) and include it in the doc on the same page as the one you just created ? 
[DIIID.pdf](https://github.com/ToFuProject/tofu/files/6733438/DIIID.pdf)

Didou09:
P.S.: I'm introducing tofu to the PlasmaPy community tomorrow afternoon, during a 'lghtning talk ' (5-10 min), I would like to show that kind of nice functionalities,
Question: do you want to do the presentation ?  (I can still prepare the slides)

lasofivec:
Yes I can do the video. Could you send me a `svg` of what you would want to have as a result and maybe the r0, z0 ? maybe the resulting config plot so I can try to recreate it

Didou09:
I don't know yet what the result should be, 
Can you do a try or two trying to get a reasonable representation of the figure ?
It's also an opportunity to figure out what functionality ay be missing to make users' life easier.

It would be nice to get:
* all PF coils
* The double vessel (grey hash on the figure), as 2 vessels
* The innermost vessel (what looks like a unique first wall approximation), as a third vessel

PR #527, opened 2021-06-30
Issue#75 sa triangle

lasofivec:
sa_map_poly0_blockFalse_steprz0_01


Didou09:
Shall we wait for this PR before the deployment of 1.4.11?
It has to be done before 13h today, what do you think?

Didou09:
Hum... the tests for the 1.4.11 are still failing because of openadas, so it looks like you have more time

tofu/geom/_GG.pyx
```diff
@@ -4932,3 +4943,481 @@ def compute_solid_angle_map(double[:,::1] part_coords, double[::1] part_r,
...
+    for ii in range(npoly):
+        temp = poly_coords[ii]
+        data[ii] = &temp[0, 0]
+    _vt.triangulate_polys(
+        &data[0],
+        &poly_lnvert[0],
+        npoly,
+        ltri,
+        num_threads
+    )
```
Didou09, line 5176: @lasofivec, here the routine  _vt.triangulate_polys() is called with the first argument as `&data[0]`
Here data is defined as `data = <double **> malloc(npoly * sizeof(double *))`

But in line 2812 (in routine `def vignetting()`, the same routine is called with the first argument as `data`.
And there data is defined as `data = <double **> malloc(nvign*sizeof(double *))`

So in both cases, data refers a similar object.
But in here it is provided as a memory address of the first element (`&data[0]`), while in line 2812 it is provided directly (`data`)

Questions:
* Why this difference ?
* Is it a bug ? 
* Which one is correct ?

The definition of the routine, in `_vignetting_tools.pyx `(line 291) says it should have a double pointer of type `double**` as first argument. Which leads to think that only the first case (here, line 5170 is correct).

More questions:
* Why use a double pointer instead of a single pointer in the routine definition ?
* Why provide a pointer only to the first element to data ?

tofu/geom/_vignetting_tools.pyx
```diff
@@ -26,24 +26,33 @@ from . cimport _sorted_set as _ss
...
     cdef double sumc
     cdef double[3] ucrossv
     # ...
-    _bgt.compute_cross_prod(u, v, ucrossv)
-    sumc = 0.0
-    for ii in range(3):
-        # normally it should be a sum, but it is a minus cause is we have (U,-V)
-        sumc = ucrossv[ii]
-    return sumc >= 0.
+    _bgt.compute_cross_prod(u, v, &ucrossv[0])
```
Didou09, line 46: why give a pointer (`&ucrossv[0]`) ? The routine takes an array (see definition)

lasofivec: To be sure it will take the actual array and not a copy of it. So that the array is modified on exit. Not really necessary but some people prefer to do it to make sure everything is modified properly.
And an array is a pointer.

Issue #540, opened 2021-07-23
Take into account constraint on wavelength interval in johann error maps 


Didou09:
Motivation:
-------------
* The current verison deos not take into account the constraint on the wavelength interval of interest
* The maps do not include that constraint => they should

Proposition:
--------------
* modify the `Crystal.calc_johannerror()` method to allow the user to specify a wavelength interval
* return a `bool` (`True`/`False`) to indicate whether the condition is satisfied or not
* Use this in the method to compute johann error map to plot the domain in which the condition is satisfied, on top of the current map

PR #542, opened 2021-07-27
Issue 540 - Wavelength Interval

adriendaros:
Motivation:
-------------
* see Issue #540 

Main changes:
----------------
* The user can determine a wavelength interval of validity thanks to new entry agruments into `cryst.plot_focal_error_summed()`.
* This condition is fullfilled into cryst.calc_johannerror() and the domain of validity is plotted as the same time of the method.
* Added possibility to plot ray-tracing in order to visualize non focalization.

Issues:
-------
* Fixes, in devel, Issue #540 

Example:
---------
```
In [1]: import tofu as tf                                                                                                                                                                                       

In [2]: import inputs_temp.XICS_allshots_C34 as xics                                                                                                                                                            

In [3]: cryst = tf.load('inputs_temp/TFG_CrystalBragg_ExpWEST_DgXICS_ArXVII_sh00000_Vers1.4.9-315-gc93fa7a3.npz')                                                                                               
Loaded from:
    /Home/AD265925/project_git/tofu/inputs_temp/TFG_CrystalBragg_ExpWEST_DgXICS_ArXVII_sh00000_Vers1.4.9-315-gc93fa7a3.npz

In [4]: det = dict(np.load('inputs_temp/det37_CTVD_incC4_New.npz', allow_pickle=True))                                                                                                                          

In [5]: xi, xj = xics._XI, xics._XJ                                                                                                                                                                             

In [6]: dax = cryst.plot_focal_error_summed(xi=xi[::10], xj=xj[::10], err='rel', plot_dets=True, det_ref=det, pts=pts)
```

image
image

### DIFF
--- a/tofu/geom/_core_optics.py
+++ b/tofu/geom/_core_optics.py
@@ -1888,11 +1888,10 @@ class CrystalBragg(utils.ToFuObject):
         bragg, phi, lamb = bragg[..., 0], phi[..., 0], lamb[..., 0]
 
         # Check lambda interval into lamb array
-        c0= (
+        c0 = (
             np.min(lamb) < lambda_interval_min
             and np.max(lamb) > lambda_interval_max
         )
-        test_lambda_interv = None
         if c0:
             test_lambda_interv = True
         else:
@@ -2085,7 +2084,7 @@ class CrystalBragg(utils.ToFuObject):
         ddist = np.linspace(dist_min, dist_max, int(ndist))
         di = np.linspace(di_min, di_max, int(ndi))
         error_lambda = np.full((di.size, ddist.size), np.nan)
-        test_lamb_interv = np.full((di.size, ddist.size), np.nan, dtype='bool')
+        test_lamb_interv = np.zeros((di.size, ddist.size), dtype='bool')
         end = '\r'
         for ii in range(ddist.size):
             for jj in range(di.size):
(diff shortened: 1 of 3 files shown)


tofu/geom/_core_optics.py
```diff
@@ -1882,6 +1887,17 @@ def calc_johannerror(
         # Only one summit was selected
         bragg, phi, lamb = bragg[..., 0], phi[..., 0], lamb[..., 0]
 
+        # Check lambda interval into lamb array
+        c0= (
+            np.min(lamb) < lambda_interval_min
+            and np.max(lamb) > lambda_interval_max
+        )
+        test_lambda_interv = None
```
Didou09, line 1895: Non necessary to initialize the variable because it will be defined in all subsequent cases

tofu/geom/_core_optics.py
```diff
@@ -2053,6 +2085,7 @@ def plot_focal_error_summed(
         ddist = np.linspace(dist_min, dist_max, int(ndist))
         di = np.linspace(di_min, di_max, int(ndi))
         error_lambda = np.full((di.size, ddist.size), np.nan)
+        test_lamb_interv = np.full((di.size, ddist.size), np.nan, dtype='bool')
```
Didou09, line 2088: There is a contradiction here,
if you initialize the array with nan => type = float
But you want type = bool

=> initialize with False, or even shorter :
`test_lamb_interv = np.zeors((di.size, ddist.size), dtype='bool')`

tofu/geom/_core_optics.py
```diff
@@ -2090,9 +2123,19 @@ def plot_focal_error_summed(
...
                     )[0],
                 )
+                test_lamb_interv[jj, ii] = self.calc_johannerror(
+                    xi=xi, xj=xj,
+                    det=det,
+                    err=err,
+                    lambda_interval_min=lambda_interval_min,
+                    lambda_interval_max=lambda_interval_max,
+                    plot=False,
+                )[4]
```
Didou09, line 2138: Good but:

* here you call the method twice (first to compute the mean erro, then to get the interval bool).
* Executing the method is the longest task => call it once, store the result to extract the 2 quantities at once

```
error_lambda_temp, test_lamb_interv[jj, ii] = 
                    self.calc_johannerror(
                        xi=xi, xj=xj,
                        det=det,
                        err=err,
                        lambda_interval_min=lambda_interval_min,
                        lambda_interval_max=lambda_interval_max,
                        plot=False,
                    )[::4]
     error_lambda[ii, jj] = np.nanmean(error_lambda_temp)
```


This syntax (`[::5]`) means 'take one every 5 indices":
So if you want to take only the first and last output, you can use `[::n]`, where `n = len(list) - 1`

```
In [5]: ll = ['a', 'b', 'c', 'd', 'e']                                                                                                               

In [6]: ll[::2]                                                                                                                                      
Out[6]: ['a', 'c', 'e']

In [7]: ll[::3]                                                                                                                                      
Out[7]: ['a', 'd']

In [8]: ll[::4]                                                                                                                                      
Out[8]: ['a', 'e']

```

Didou09 (approved):
I approve this PR

It is clear

I have a few minor comments / suggestions to imrpove a bit the code (see below)
When they are implemented you can merge the PR

You must of course make sure all tests are passing before merging, but once the tests are passing, you can mere

Issue #555, opened 2021-10-06
Boost _GG.compute_solid_angle_map()


Didou09:
Motivations:
--------------
* @obrejank is going to dig into the solid angles to help and learn from @lasofivec 
* Before developping a new function from scratch, a first stpe to understand Laura's code is to try and see if he can optimize a small part of it.
* As a first try, and as a learning exercice, he will try to see if he can (or ot ?) optimize a bit `_GG.compute_solid_angle_map()`
* Since we have a meeting in one week, it will give him the opportunity to get familiar with the code, try some tricks and ask questions to @lasofivec on Friday.

Proposal:
--------------
* Create a dedicated branch `Issue555_BoostSolidAngleMap`, from devel, and work on it
* Beware of : 
    * parallelisation : it is done with openmp, possibly several nested loops parallelized
    * the routine samples the tokamak volume, computes the solid angle subtended by a spherical particle for each point, and integrates torodally to produce a R, Z map of the integrated solid angle. You will quickly realize that the volume sampling is not done on a cylindrical coordinates basis. Instead it is done with the constraint that the volume elements `dR * Rdphi * dZ` must be constant, so at 2 given `R` correspond 2 different numbers of toroidal elements.
    * The routine can run for several particles, each associated to a position (X, Y, Z) and a radius.

How to run it:
-------------------

Once you have cloned and branch the repo, and you are at the root of your local repo, you need to compile the library

`python setup.py build_ext --inplace`

Then you can use it and run the routine, here is a minimum working example:

```
In [1]: import tofu as tf
/home/didier/Documents/Projects/tofu/tofu/__init__.py:93: UserWarning: 
The following subpackages are not available:
	- tofu.imas2tofu
	- tofu.mag
  => see print(tofu.dsub[<subpackage>]) for details.
  warnings.warn(msg)

In [2]: conf = tf.load_config('WEST')

n [3]: ptsXYZ = np.array([[2.4, 2.5, 2.6], [0., 0., 0.], [0., 0., 0.]])

In [4]: radius = np.r_[100.e-6, 100.e-6, 100.e-6]

In [5]: RZ, sang, ind, reseff, dax = conf.calc_solidangle_particle_integrated(part_traj=ptsXYZ, part_radius=radius, resolution=0.1, DR=None, DZ=None, DPhi=[-np.pi/2, np.pi/2])

```

[...]

PR #535, opened 2021-07-15
[Issue529] Created Config.calc_solidangle_particle_integrated()

Didou09:
Motivation:
------------
* @lasofivec implemented a low-level parallelized function to compute the toroidal integral of the solid angle subtened by a 491spherical particle from all points in the torus (cf. PR #491 )
* A new method is added to Config to use this new function, to make it easy for the user, who can choose the spatial resolution, and get a plot of the result
* Units tests are updated to test this method
* This new method is quite easy to handle and the plotting can be used for debugging

Issues:
-------
* This fixes, in devel, issue #529 
* A new issue may have to be created due to strange results obtained in some cases

Example:
----------

```
In [1]: import tofu as tf                                                                                                                                                              
conf = /Home/DV226270/ToFu_All/tofu_git/tofu/tofu/imas2tofu/__init__.py:87: UserWarning: 
You do not seem to be using the latest IMAS version:
'module list' vs 'module av IMAS' suggests:
	- Current version: 3.30.0-4.8.4
	- Latest version : 3.32.1-4.9.1-R2019b
  warnings.warn(msg)

In [2]: conf = tf.load_config('AUG', strict=True)                                                                                                                                      

In [3]: part_radius = np.array([10e-6, 10e-6, 10e-6])                                                                                                                                  

In [4]: part_traj = np.array([[1.8, 0.01, 0.], [1.8, 0.01, -0.1], [1.8, 0.01, -1]]).T                                                                                                                                                           

In [5]: out = conf.calc_solidangle_particle_integrated(part_traj=part_traj, part_radius=part_radius, resolution=0.05, vmax=False)     

```

image

What seems wrong in the image above:
in the left plot:
* the particle seem to be cenetred around Z=-0.1, which is wrong, it is around Z = 0 (it should be on the red line)
* Moreover, the integrated solid angle should be more or less symetric as we go away from the particle, but here there is clear a brutal decrease as we go from the particle towards Z<0 (red cirlce)

=> Opening a dedicated issue

PR #578, opened 2021-11-12
Issue555 boost solid angle map

obrejank:
This branch introduces a few optimisations and OMP parallelisation to `sa_assemble_arrays` and its 4 computation methods: 
- `sa_assemble_arrays` now pre-computes the pairs `(rr, zz)` such that `is_in_vignette[rr, zz]==true` and other arrays (such as `pts_mv`) to improve parallel work balance and reduce code redundancy in the 4 methods, 
- `assemble_block_approx` and `assemble_block_exact` were parallelised,
- `assemble_unblock_approx` and `assemble_unblock_exact` were already parallelised but now use the schedule `guided`.

The figure below summarises the scaling improvements, by scanning over `OMP_NUM_THREADS` in the test code mentioned in #555 in which the flags were changed to call all 4 methods, the resolution adjusted to obtain a suitable computation time (the time reported were ontained by adding timers inside `sa_assemble_arrays` and it excludes the unparallelised python overhead added when calling `conf.calc_solidangle_particle_integrated`). The benchmark was performed on a computer with 20 (physical) cores, and the reduced gain at 20 cores is most likely due to competition for resources with other programs. 
scaling
(For the unblocked methods, `ref` corresponds to the original parallelisation)


Didou09:
Thank you @obrejank !

So:
* A quick look at the code changes seems ok to me, I don't scrupulously check each line but I get the point of most modifications and they seem understandable to me
* Re-running the example given in the PR #535 that produced the original version of this code seems to give the same result, which is good news (see below). Here I have the colorbar max to 1e-8 in both cases to try and visualize any difference.

New version:
image

Original version:
image


* Running the new standardized benchmark on both the original version and this PR reveals a very good 82 % speed gain in the particular case tested in the benchmark, which is great !

image

Hence, it seems everything is fine and well done,
Congratulations and thanks again,

Thanks also @lasofivec for writing the original version and doing the whole work of writing the structure and algorithms and @flothesof for suggesting that idea of standardized benchmarks using [asv](https://asv.readthedocs.io/en/stable/index.html), they are already useful !

Validating and merging

Didou09:
Branch deleted

Issue #582, opened 2021-11-15
Optional svg.path dependency


flothesof:
Hi @Didou09 

following release of 1.5.0, all tests except one pass in the conda-forge build.
See e.g. https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=411920&view=logs&j=4f922444-fdfe-5dcf-b824-02f86439ef14&t=b2a8456a-fb11-5506-ca32-5ccd32538dc0

This is due to the svg.path dependency.
Should I add this dependency to the build so that everything passes?

Cheers,
Florian

PR #583, opened 2021-11-15
[#582] Added svg.path to setup.py, requirements.txt, meta.yaml

Didou09:
Motivations:
--------------
* see issue #582 
* Adding svg.path as a compulsory dependency for users results from the following reasons:
    * to avoid crashing conda-forge
    * because importing a tokamak geometry from inkscape is one of the popular features of tofu
    * because svg.path seems available on ll platforms both from [Pypi](https://pypi.org/project/svg.path/) and [anaconda.org](https://anaconda.org/conda-forge/svg.path)

Main changes:
----------------
* Added svg.path to setup.py, requirements.txt, conda_recipe/meta.yaml

### DIFF
--- a/.github/workflows/python-package-pip-testing.yml
+++ b/.github/workflows/python-package-pip-testing.yml
@@ -28,6 +28,7 @@ jobs:
         echo $CONDA/bin >> $GITHUB_PATH
     - name: Install dependencies
       run: |
+        conda config --append channels conda-forge
         conda env update --file requirements.txt --name base
     - name: Lint with flake8
       run: |
--- a/conda_recipe/meta.yaml
+++ b/conda_recipe/meta.yaml
@@ -41,6 +41,7 @@ requirements:
     - pytest
     - pygments
     - requests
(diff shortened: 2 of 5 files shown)


Didou09:
Hum.... bad news, it seems that adding the svg.path dependency to the requirements.txt does not work:

image

We'll need a bit more thinking through this...
Let me know your thoughts @flothesof , @lasofivec :-)

Didou09:
I may have found why: the particular github workflow called `python-package-pip-testing.yml` builds an environment using `conda` with no default channels set.
Other workflows build their environments using `pip`.

As it turns out, `svg.path` exists on `Pypi` (for `pip`) but not on the default channel of `conda`, only on `conda-forge` (not included by default).

I've added `conda-forge` to the channels to be used by `python-package-pip-testing.yml` and it seems to do the trick (tests still running but nothing failed as quickly as previously).

lasofivec:
Yes, all other required packages are on the default channel

PR #590, opened 2021-11-17
Taking out cpp dependency

lasofivec:
I recently came across this bug on my machine:

`undefined symbol: _ZSt28__throw_bad_array_new_lengthv`

which seems to have been seen in different projects

https://github.com/conda/conda/issues/10757
https://github.com/cython/cython/issues/4218
https://github.com/JuliaLang/julia/issues/40703

Only workarounds have been proposed and since it something that I has only been seen with the latest gcc, I didn't want to fix the problem only on my machine. 
Checking a little bit more into detail, it looked like the problem only came when using the `vector` and `set` classes of `c++`. Since these are the only two instances where c++ was needed and that we only need them for some specific cases, I decided to implement a minimalist version of those classes for the use of tofu.

`ChainedList` (replacing vector) and `SortedSet` (replacing `set`), only have a couple of functions in order to lighten the code.


Didou09 (approved):
Fine for me, this looks clean and working, thanks

Issue #599, opened 2021-12-09
Issues with linux installation


fsciortino:
Following the `installing ToFu as a developer` section in https://tofuproject.github.io/tofu/installation.html, I run into some trouble. I tested this on 2 different servers, both with and without conda. Specifically, the error occurs after running `pip install -e .[dev]`. Since I don't have root access on either of the 2 servers that I tested on, in practice I run `python3 -m pip install -e . --user`.

The issue seems to be with the installation trying to write to root-owned directories. The part of the output that seems relevant for debugging, coming after the installation of all dependencies, is the following (here, on the ltxs cluster at IPP-Garching):

```
  Running setup.py develop for tofu
    ERROR: Command errored out with exit status 1:
     command: /usr/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/afs/ipp-garching.mpg.de/home/s/sciof/tofu/setup.py'"'"'; __file__='"'"'/afs/ipp-garching.mpg.de/home/s/sciof/tofu/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps --user --prefix=
         cwd: /afs/ipp-garching.mpg.de/home/s/sciof/tofu/
    Complete output (43 lines):
    
    Version for setup.py :  1.5.0-61-g2e6e9523
    
    running develop
    WARNING: The user site-packages directory is disabled.
    /usr/lib/python3.6/distutils/extension.py:131: UserWarning: Unknown Extension options: 'language_level'
      warnings.warn(msg)
    /usr/lib/python3.6/distutils/dist.py:261: UserWarning: Unknown distribution option: 'use_scm_version'
      warnings.warn(msg)
    /tmp/pip-build-env-vc1b3n_u/overlay/lib/python3.6/site-packages/setuptools/dist.py:508: UserWarning: The version specified ('1.5.0-61-g2e6e9523') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
      "details." % version
    /tmp/pip-build-env-vc1b3n_u/overlay/lib/python3.6/site-packages/setuptools/command/easy_install.py:159: EasyInstallDeprecationWarning: easy_install command is deprecated. Use build and pip and other standards-based tools.
[...]

PR #609, opened 2022-01-13
update doc as of 2022

flothesof:
As discussed at the end of https://github.com/ToFuProject/tofu/issues/599#issuecomment-990032743, here is a small review and improvement of the docs regarding mostly installation.

Notable changes to doc:
- removed specialized instructions in favor of conda-forge for normal users: I believe it’s simpler to advertise this "one size fits all solution" rather than the previous version
- rewrote developer install and made sure it works on my machine
- better links to other parts of the doc

Possible improvements I saw while working on this:
- the "normal" tests took 11 minutes to complete, this is really crazy since they took less than 2 minutes a couple of months ago --> this is something that will probably deter contributors
- I don’t understand the "how to construct the doc" section https://tofuproject.github.io/tofu/dev_tutos.html#how-to-construct-the-doc, in my opinion it would be worth reworking @lasofivec what do you think?
- I haven’t checked the README, it would probably be a good idea to check it for coherence with these install instructions

@Didou09 @lasofivec what do you think?

### DIFF
--- a/doc/source/contributing.rst
+++ b/doc/source/contributing.rst
@@ -1,3 +1,5 @@
+.. _contributing-to-tofu:
+
 Contribute to tofu
 ==================
 
--- a/doc/source/dev_tutos.rst
+++ b/doc/source/dev_tutos.rst
@@ -74,8 +74,9 @@ Adding text, updating the API:
 #. Commit and push the changes to the ``new-doc`` branch: to keep the changes in the sources
 #. Commit and push the changes to the ``gh-pages``, this will update the website.
 
-
-Adding a tutorial:
+-----------------
+Adding a tutorial
+-----------------
 
 #. Add your tutorial in ``$TOFU_DIR/examples/tutorials`` with the proper naming conventions
 #. The sphinx gallery is automatically updated when building the doc (as above)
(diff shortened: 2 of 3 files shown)


Didou09 (approved):
I am ok with this PR, thanks you @flothesof

Didou09:
On the additional thougts:
* **tests duration:** I agree 11 minutes is bit too long, will try to reduce that (2 commits in PR # already address it partially, but more can probably be done in the near-future)
* **how to construct the doc:** I'll try to go through it again
* **README**: will try to go through it before the next release


Thanks !

lasofivec:
> I don’t understand the "how to construct the doc" section https://tofuproject.github.io/tofu/dev_tutos.html#how-to-construct-the-doc, in my opinion it would be worth reworking @lasofivec what do you think?

Hi Flo, we went over it not so long ago with Didier. What did you find confusing? Could you elaborate on that?

flothesof:
@lasofivec Sure: 
- the title says "how to construct the doc" but the content describes two things: building + uploading to GH (if I understand correctly)
- I don’t understand why a two repo solution is needed (in my experience, I can build the doc in the tofu directory), I find it rather confusing (but see above point: maybe I misunderstood the goal of the section)
- under windows I couldn’t get make html working (and ran the build with sphinx-build instead)

Issue #629, opened 2022-02-10
Add non-parallelism into crystal for rocking curve computations


adriendaros:
* From the recently wrote method `cryst.compute_rockingcurve()` computing rocking curves for a alpha-Quartz crystal with specified Miller indices `(h,k,l)` and wavelength `lamb`, create new lines permitting to add non-parallelism.
* This non-parallelism, represented by an angle `alpha`, affects the general equations used to compute the reflectivity of the crystal. Then, power-ratio curves have to be plotted for each `alpha` angle and also the integrated reflectivity.
* This task is the final step of translating work of Francesca's code.

PR #630, opened 2022-02-25
Issue629 non parallelism into rocking curves

adriendaros:
Motivations:
---------------
* Integer an asymmetry property called non-parallelism between crystal's optical surface and inter-atomic planes into the rocking curve computations.
* Study the effect of the well-known non-parallelism specific of Bragg crystals used in WEST on their reflecting curves. 

Results:
----------
* This option is now available through the entry argument `use_non_parallelism = True`.
* This asymmetry is designed to vary between plus and minus the Bragg angle corresponding to the chosen wavelength, within 0.01 radians.
* The argument `na` let the choice of the number of steps between the asymmetry limits. It fixed to 51 steps by default.
* The `na` values of the integrated reflectivities, ratio between parallel and perpendicular components, widths of curves and non-parallelism angles are then printed.

Issues:
--------
* Fixes, in devel, issue #629 

Examples:
------------
image
image
image
image

### DIFF
--- a/tofu/geom/_core_optics.py
+++ b/tofu/geom/_core_optics.py
@@ -803,12 +803,16 @@ class CrystalBragg(utils.ToFuObject):
 
     def compute_rockingcurve(
         self, ih=None, ik=None, il=None, lamb=None,
-        plot_asf=None, plot_power_ratio=None,
+        use_non_parallelism=None,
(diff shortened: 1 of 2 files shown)


Didou09 (approved):
I approve this PR

This is good and clean work, I did not find anything that would justify a change,
well done !

Issue #663, opened 2022-07-28
Optimize solid angle computation - new


Didou09:
Context:
-----------

The routines mentioned in issue #653 have been written, the are functional and have unit tests.

They would need to be profiled / benchmarked / optimized / parallelized:

A benchmark has been prepared and can be run with:

```
(base) didier@didier-XH310V2~/Documents/Projects/tofu (Issue658_Etendue) $ asv run HEAD^! -q -b benchmarks_03_solidangles -e
· Creating environments
· Discovering benchmarks
· Running 4 total benchmarks (1 commits * 1 environments * 4 benchmarks)
[  0.00%] · For tofu commit 1b83950e <Issue658_Etendue>:
[  0.00%] ·· Benchmarking conda-py3.8.12-cython0.29.24-matplotlib3.4.3-numpy1.21.2-requests2.26.0-scikit-sparse-scikit-umfpack-scipy1.7.1
[  0.00%] ··· Importing benchmark suite produced output:
[  0.00%] ···· /home/didier/Documents/Projects/tofu/.asv/env/046138848c127820e353d55c8576a687/lib/python3.8/site-packages/numpy/core/fromnumeric.py:3208: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
                 return asarray(a).size
               /home/didier/Documents/Projects/tofu/.asv/env/046138848c127820e353d55c8576a687/lib/python3.8/site-packages/tofu/__init__.py:98: UserWarning: 
               The following subpackages are not available:
               	- tofu.imas2tofu
               	- tofu.mag
               	- tofu.tomotok2tofu
                 => see print(tofu.dsub[<subpackage>]) for details.
                 warnings.warn(msg)

[ 12.50%] ··· Setting up benchmarks_03_solidangles.py:46                                                                                                                                                    ok
[ 12.50%] ··· benchmarks_03_solidangles.SolidAngle.time_00_light                                                                                                                                       961±0ms
[ 25.00%] ··· benchmarks_03_solidangles.SolidAngle.time_01_unitvectors                                                                                                                                 978±0ms
[ 37.50%] ··· benchmarks_03_solidangles.SolidAngle.time_03_etendue_check                                                                                                                              49.0±0ms
[...]

PR #715, opened 2022-12-02
Optimisation of the solid angle computation with aperture obstruction

obrejank:
This MR is a follow up to #663 providing major performance improvement to all to the 3 functions `compute_solid_angle_apertures_light` (which have also all been merged into one) as well as OMP parallelisation. In my tests, the speedup was about 60x in sequential (`OMP_NUM_THREADS=1`). The computation seems to be mostly memory-bound so the parallelisation efficiency is not very large (x3 with 8 threads), but it is still better than nothing and I'm not sure much more can be gained.

### DIFF
--- a/tofu/geom/_GG.pyx
+++ b/tofu/geom/_GG.pyx
@@ -44,6 +44,379 @@ from . cimport _sampling_tools as _st
 from . cimport _vignetting_tools as _vt
 from . import _openmp_tools as _ompt
 
+from libc.stdio cimport printf
+from openmp cimport omp_get_wtime, omp_get_thread_num, omp_get_max_threads
+
+# == Polygon intersection ======================================================
+
+ctypedef struct polygon:
+    int size
+    double * vx
+    double * vy
+    polygon * next_component
+
+ctypedef struct polygon_node:
+    double x, y
+    polygon_node *next
+    polygon_node *prev
+    polygon_node *neighbour
+    bint entry
+    bint visited
+    double alpha
+
+cdef polygon_node * polygon_node_new(
+        double x,
+        double y,
+        polygon_node * prev=NULL,
+        polygon_node * next=NULL ) nogil:
+
+    cdef polygon_node * p
+    p = <polygon_node *>malloc(sizeof(polygon_node))
+    p.x = x
+    p.y = y
+    p.prev = prev
+    p.next = next
+    p.neighbour = NULL
+    p.visited = False
+
+    if prev != NULL:
+        p.prev.next = p
+    if next != NULL:
+        p.next.prev = p
+
+    return p
+
+cdef polygon_node * polygon_node_next_vertex( polygon_node * p ) nogil:
+
+    while p.next.neighbour != NULL:
+        p = p.next
+
+    return p.next
+
+cdef polygon_node * polygon_node_next_intersec( polygon_node * p ) nogil:
+
+    while p.next.neighbour == NULL:
+        p = p.next
+
+    return p.next
+
+cdef void polygon_node_insert_intersec(
+        double x, double y,
+        polygon_node * poly1_node, double alpha1,
+        polygon_node * poly2_node, double alpha2 ) nogil:
+
+    while poly1_node.next.neighbour != NULL and poly1_node.next.alpha < alpha1:
+        poly1_node = poly1_node.next
+    poly1_node = polygon_node_new(x, y, poly1_node, poly1_node.next)
+
+    while poly2_node.next.neighbour != NULL and poly2_node.next.alpha < alpha2:
+        poly2_node = poly2_node.next
+    poly2_node = polygon_node_new(x, y, poly2_node, poly2_node.next)
+
+    poly1_node.alpha = alpha1
+    poly1_node.neighbour = poly2_node
+    poly2_node.alpha = alpha2
+    poly2_node.neighbour = poly1_node
+
+cdef bint polygon_node_is_in(
+        polygon_node * poly_point,
+        polygon_node * poly ) nogil:
+
+    cdef polygon_node * p0
+    cdef polygon_node * p1
+    cdef double x, y
+    cdef double slope
+    cdef bint ret
+
+    x = poly_point.x
+    y = poly_point.y
+
+    if poly.neighbour != NULL:
+        poly = polygon_node_next_vertex(poly)
+
+    ret = False
+    p0 = NULL
+    p1 = poly
+    while True:
+        p0 = p1
+        p1 = polygon_node_next_vertex(p1)
+
+        if p1.x == x and p1.y == y:
+            return True
+        if (p0.y > y) != (p1.y > y): # xor
+            slope = (
+                (x-p1.x)*(p0.y-p1.y)
+                - (p0.x-p1.x)*(y-p1.y)
+            )
+            if slope == 0.: # point is on a vertex
+                return True
+            if (slope < 0.) != (p0.y < p1.y):
+                ret = not ret
+
+        if p1 == poly:
+            break
+
+    return ret
+
+cdef void polygon_node_free_all(polygon_node * p) nogil:
+    p.prev.next = NULL
+    while p.next != NULL:
+        p = p.next
+        free(p.prev)
+
+cdef void polygon_intersect(polygon * poly1, polygon * poly2, polygon * res, bint ignore_duplicates=False) nogil:
+    cdef int i, i1, i2
+    cdef int intersection_count
+    cdef double perturbation, epsx, epsy
+    cdef double u1x, u1y, u2x, u2y
+    cdef double dpx, dpy
+    cdef double D, alpha1, alpha2
+    cdef double px, py
+    cdef polygon_node * pn
+    cdef polygon_node * poly1_node
+    cdef polygon_node * poly2_node
+    cdef polygon_node * poly1_node_next
+    cdef polygon_node * poly2_node_next
+    cdef bint entry
+
+    # ----------------------
+    # Empty res if it is not
+    if res.size != 0:
+        res.size = 0
+        free(res.vx)
+        free(res.vy)
+        res.vx = NULL
+        res.vy = NULL
+
+    # ----------------------
+    # Avoid degenerate cases
+    if poly1.size<=2 or poly2.size<=2:
+        return
+
+    if poly1.next_component != NULL or poly2.next_component != NULL:
+        printf("[WARNING] polygon_intersect: disconnected polygons not implemented yet; ignoring other components\n")
+        return
+
+    # ----------------------------------------------
+    # Phase 0: creating polygon_node representations
+    # Perturbation of polygon 1 to evacuate problematic cases
+
+    cdef double xmin1 = poly1.vx[0]
+    cdef double xmax1 = poly1.vx[0]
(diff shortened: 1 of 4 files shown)


Didou09:
Thank you very much @obrejank , I'll fix the unit tests (`matplitlub._contour`), and it should work soon !

Didou09:
OK, the workflow is fixed, 

But, when running the full workflow (a 3x3 matrix with python `3.7`, `3.8`, `3.9` on `Ubuntu`, `Windows` and `MacOS`), it appears that the installation of `tofu` does not work anymore on `MacOS`:

https://github.com/ToFuProject/tofu/actions/runs/3620729344/jobs/6103424271

The error message suggests an issue with the compilation of `tofu` with `OpenMP`.

I have checked the following :

- [ ] I have temporarily removed testing on `MacOS` to check that the unit tests are all passing on `Ubuntu` and `Windows`
- [ ] I have checked that the devel branch still works on `MacOS` => this issue is specific to this branch and is not due to any change on the MacOS platforms themselves.

Hence, this issue is specific to MacOS and specific to this branch.

Any chance you could take a closer look into this MacOS-specifc issue @obrejank ?
Is there anything that you have modified that could have changed the way tofu compiles on MacOS ?

To edit the file that runs the unit tests on the testing matrix, you can play with the following file:
`.github/workflows/complete-matrix.yml`

obrejank:
I only had a quick look, but the final error messages says:
` error: Command "gcc -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -I/usr/local/opt/sqlite/include -I/usr/local/opt/sqlite/include -I/Users/runner/hostedtoolcache/Python/3.7.15/x64/lib/python3.7/site-packages/numpy/core/include -I/Users/runner/hostedtoolcache/Python/3.7.15/x64/include/python3.7m -I/Users/runner/hostedtoolcache/Python/3.7.15/x64/include/python3.7m -c tofu/geom/_GG.c -o build/temp.macosx-10.15-x86_64-3.7/tofu/geom/_GG.o -O3 -Wall -fno-wrapv" failed with exit status 1`
where the last few flags are the "extra flags" defined in `setup.py`, which should include the `-fopenmp` flag returned by `is_openmp_installed`. From what I understand, it can only return nothing if the `try except finally` clause failed, but I don't have a Mac to test it on and see why the `subprocess.call` failed.

obrejank:
If you look at the output in https://github.com/ToFuProject/tofu/actions/runs/3622194828/jobs/6106600138, you will see that MacOS uses clang when you call gcc:
`subprocess.call(..) -> CompletedProcess(args=['gcc', '-fopenmp', 'test.c'], returncode=1, stdout=b'', stderr=b"clang: error: unsupported option '-fopenmp'\nclang: error: unsupported option '-fopenmp'\n")`
which is absurd and stupid, but confirmed [here](https://stackoverflow.com/questions/33357029/using-openmp-with-clang). Since nothing was changed in recent versions of ToFu, I assume that this comes from an update in github's VMs.

Didou09:
A possible fix could be to drop temporarily the support for `MacOS`, but several users are using `MacOS`.... It has to be fixed at some point

Didou09:
So, I tried providing the correct argument:
image


But it fails because the default CLANG install on MacOS does not support OpenMP:

https://github.com/ToFuProject/tofu/actions/runs/3622194828/jobs/6106600138

` subprocess.call(..) -> CompletedProcess(args=['gcc', '-fopenmp=libomp', 'test.c'], returncode=1, stdout=b'', stderr=b"clang: error: unsupported argument 'libomp' to option 'fopenmp='\nclang: error: unsupported argument `

https://www.positioniseverything.net/clang-error-unsupported-option-fopenmp

And apparently we have to make sure it's `llvm` that's installed and not just the default clang

What I still don't understand is why it works on other branches...

Didou09:
Found llvm here:

https://github.com/kylemayes/install-llvm-action

Didou09:
Identified source of compiling issue:
-----------------------------------------------
* In the prevous versions, `openmp` was an optional dependency, indeed the only module that was explicitly doing a `cimport openmp` was `./tofu/geom/_openmp_tools.pyx`, but it was a conditional import

image

The rest of the module is coded with a condition on the existence of `openmp`:
image

* On this branch, `_GG.pyx` now has an unconditional `cimport openmp`

So basically, the compiling issue we have here was probably existing already in the previous version, but it was not apparent because for MacOS `TOFU_OPENMP_ENABLED` was `False` and MacOS tests were run without parallelization.

We have 2 options:
1. Implement the test on `TOFU_OPENMP_ENABLED` in `_GG.pyx `(=> lots of extra work)
2. Make sure `openmp` works on MacOS => better but difficult

Didou09:
I was testing this a bit more extensively and found a bug that can be reproduced with the MWE script attached.

It creates a set of sensors, with 2 apertures each.
Each sensors is bigger than the previous.
The resulting solid angle map should increase accordingly (on devel).

On this branch, the solid angle map becomes degenerate when the sensor becomes big, and even decreases instead of increasing, for some reason.
It may look like a rounding error?
[NCAM_tutorial_tofu.txt](https://github.com/ToFuProject/tofu/files/12637147/NCAM_tutorial_tofu.txt)

```
import numpy as np
plt.switch_backend('Qt5Agg')
plt.ion()

import tofu as tf


conf = tf.load_config('SPARC-V0')

# --------------------------------
# Instanciate an empty Collection

coll = tf.data.Collection()

# add mesh
coll.add_mesh_2d_rect(
    key='m0',
    res=0.05,
    domain=[(1.3, 2.4), (-1, 1)],
    crop_poly=conf,
)



# ----------------------
# use the built-in method

coll.add_camera_pinhole(
    #keys / identifiers
    key='cam0',
    key_pinhole='pinhole0',
    key_diag='ncam0',
    cam_type='1d',
    # coordinates of the pinhole (x, y, z) or (R, Z, phi)
    R=5,
    z=0,
    phi=0,
    # 3 angles to orientate the camera
    # theta is the angle vs horizontal plane (0 => pointing towards R=inf)  in a poloidal plane
    theta=np.pi,
    # dphi is an angle vs the poloidal plane
    dphi=0.,
    # tilt is an angular tilt of the camera around its own axis
    tilt=np.pi/2.,
    # length between pinhole and sensor plane
    focal=10,
    # nb / size of pixels
    pix_nb=5,
    pix_size=0.1,
    pix_spacing=1.,
    # pinhole: circular or rectangular
    pinhole_radius=0.1,
    pinhole_size=None,
    # tokamak (Config) to be associated to it
    config=conf,
    # shall LOS computation start right away?
    compute=True,
)

# --------------------------
# extract los pts from ncam0

# pts (2, ncam)
ptsx, ptsy, ptsz = coll.get_rays_pts('ncam0_cam0_los')

# units vectors of LOS (1, ncam)
vectx, vecty, vectz = coll.get_rays_vect('ncam0_cam0_los')

# prepare sensor size
size = 0.001

# prepare theta for describing the circular outline of apertures
theta = np.pi * np.linspace(-1, 1, 50)

# prepare radius of collimator
radius = 0.02

# prepare distance between sensor and apertures
d0 = 0.2
d1 = 3.2

# ------------------
# Build the geometry

# get starting points of
doptics = {}
ncam = ptsx.shape[1]
size = size * (np.arange(ncam)*10 + 1)
for ii in range(ncam):
# for ii in range(1):

    # set keys of camera, aperture 0 and 1 (2 ends of the collimator)
    kcam = f'ncam1_c{ii}'
    kap0 = f'cam1_ap{ii}0'
    kap1 = f'cam1_ap{ii}1'

    # center of sensor
    cent = np.r_[ptsx[0, ii], ptsy[0, ii], ptsz[0, ii]]

    # los unit vector
    nin = np.r_[vectx[0, ii], vecty[0, ii], vectz[0, ii]]
    e0 = np.r_[-nin[1], nin[0], 0.]
    e0 = e0 / np.linalg.norm(e0)
    e1 = np.cross(nin, e0)

    # geometry dict for single-pixel camera 1d
    dgeom = {
        'cents_x': np.r_[cent[0]],
        'cents_y': np.r_[cent[1]],
        'cents_z': np.r_[cent[2]],
        'nin_x': np.r_[-1],
        'nin_y': np.r_[0],
        'nin_z': np.r_[0],
        'e0_x': np.r_[0],
        'e0_y': np.r_[-1],
        'e0_z': np.r_[0],
        'e1_x': np.r_[0],
        'e1_y': np.r_[0],
        'e1_z': np.r_[1],
        'outline_x0': size[ii] * np.r_[-1, 1, 1, -1],
        'outline_x1': size[ii] * np.r_[-1, -1, 1, 1],
    }

    # add single-pixel 1d camera
    coll.add_camera_1d(key=kcam, dgeom=dgeom)

    # add apertures
    dgeom = {
        'cent': cent + d0 * nin,
        'nin': nin,
        'e0': e0,
        'e1': e1,
        'outline_x0': radius * np.cos(theta),
        'outline_x1': radius * np.sin(theta),
    }
    coll.add_aperture(key=kap0, **dict(dgeom))

    # update center position
    dgeom['cent'] = cent + d1 * nin
    dgeom['outline_x0'] = radius * np.cos(theta)
    dgeom['outline_x1'] = radius * np.sin(theta)
    coll.add_aperture(key=kap1, **dict(dgeom))

    # fill in doptics
    doptics[kcam] = [kap0, kap1]


# ---------------------------------------------------------------
# create the diagnostic ncam1 using the info contained in doptics

coll.add_diagnostic(
    key='ncam1',
    doptics=doptics,
    compute=True,
    config=conf,
)


# --------------------------
# Compute vos

dvos, dref = coll.compute_diagnostic_vos(
    # which diag and mesh to use
    key_diag='ncam1',
    key_cam=None,
    key_mesh='m0',
    # sampling resolution (to be tuned by multiple tests until you find reasonable convergence)
    res_RZ=0.03,
    res_phi=0.01,
    # extra flags
    visibility=False,
    verb=None,
    store=False,
)


# --------------------
# prepare plot

fig = plt.figure(figsize=(15, 6))
ax = fig.add_axes([0.03, 0.02, 0.95, 0.95], aspect='equal')
ax.set_xlabel('R (m)', size=12, fontweight='bold')
ax.set_ylabel('Z (m)', size=12, fontweight='bold')

dax = {'cross': ax}
for ii in range(ncam):

    kcam = f'ncam1_c{ii}'

    # cam0
    dax = coll.plot_diagnostic_vos(
        key='ncam1',
        key_cam=kcam,
        indch=None,
        elements='o',
        dvos=dvos,
        # specify colormaps
        proj='cross',
        dax=dax,
        plot_config=conf,
        plot_colorbar=False,
    )

    # add detector size
    ax.text(
        np.hypot(ptsx[0, ii], ptsy[0, ii]) + 0.05,
        ptsz[0, ii],
        f'{size[ii]}',
        c='k',
        horizontalalignment='left',
        verticalalignment='center',
    )

    # add max solid angle
    sang = dvos[kcam]['sang']['data']
    ax.text(
        np.hypot(ptsx[1, ii], ptsy[1, ii]) - 0.05,
        ptsz[1, ii],
        f"{np.max(sang):3.2e}",
        c='k',
        horizontalalignment='right',
        verticalalignment='center',
    )

```



On `devel`:
NCAM_devel

On **this branch**:
NCAM_Issue663

Issue #747, opened 2023-04-14
Jumps in spectrometer data


cjperks7:
I think some bug in the spectrometer calculations was introduced since upgrading to tofu1.7.0

If I build my spectrometer and then plot it using `diag.plot_diagnostic(key='XRSHRKr', plot_config=conf, elements='o', data='XXX')`, where 'XXX' can be really any of the options, I observe discrete jumps versus pixel. Shared is a selection of these plots illustrating the error.

I recall we had a similar issue like this before, but you had fixed it and everything had been working smoothly up till now.

Minimum working code: Running the first 3 steps of the `tofu_sparc` workflow then plotting as given above.
<img width="1368" alt="Screenshot 2023-04-14 at 9 49 25 AM" src="https://user-images.githubusercontent.com/71457911/232065278-81f59d3f-2713-49f7-ae6a-71c1332094c4.png">
<img width="1419" alt="Screenshot 2023-04-14 at 9 51 18 AM" src="https://user-images.githubusercontent.com/71457911/232065296-443fc690-adf8-45d0-94de-be774135fe41.png">
<img width="1384" alt="Screenshot 2023-04-14 at 9 51 46 AM" src="https://user-images.githubusercontent.com/71457911/232065315-3011265c-1996-4a4a-bf33-b48f5dbdaf6f.png">

Didou09:
I'll have to look into it, I suspect it comes again from numerical inaccuracies in the way the projection of a polygon is handled on a concave reflexive surface (can you confirm this is a concave crystal ?)

cjperks7:
This is using a von Hamos

Didou09:
@cjperks7 I think I have identified and corrected what is causing this, 
Could you add more details on the configuration you're building so I can try to reproduce it to check it's solved ?

cjperks7:
This is running the first three steps of the tofu_sparc workflow. Here's code you can copy/paste

```
# Modules
import tofu_sparc as tfs
import os
import tofu as tf

from scipy.interpolate import interp1d
import scipy.constants as cnt
import copy

# Step00, Intializes diagnostic object and beamlines
diag, conf, dhp, key_dhp, dportref, dtube, drooms, dk, dax = tfs.step00_get_beamlines()
#diag, conf, dhp, key_dhp, dportref, dtube, drooms, dk = tfs.step00_get_beamlines()

# Step01, Determines spectrometer geometry
[...]

Issue #748, opened 2023-04-14
Minor bugs observed in compute_diagnostic_signal()


cjperks7:
I found a few things that seem like small bugs in compute_diagnostic_signal() routine

1) In `_class8_compute_signal._compute_los()` if "spectro=True", key_cam wasn't being passed into `get_diagnostic_lamb()`. See the attached screenshot. Note that I used key_cam = key_cam[0] because `compute_diagnostic_signal()` accepts key_cam (and even turns into if initially a string) as a list while `get_diagnostic_lamb()` doesn't. I do all the cameras at the same time and conveniently the wavelength basis is the same for each, but this isn't the most general fix
<img width="706" alt="Screenshot 2023-04-14 at 1 14 28 PM" src="https://user-images.githubusercontent.com/71457911/232112506-a27cdaaf-c752-4b7d-b89d-c4c7540e2194.png">

2) I've been running this function with `compute_diagnostic_signal(brightness=True)` to compare the results of each step with Tofu to the old workflow. It seems like the units for the brightness is wrong though. I input a spectral-resolved emissivity field with units ph/s/m3/sr/m whereas the units for the brightness as shown with `diag.show('data')` is in ph/s/m3/sr. Upon inspection of the code, it lost a unit of meter because of spectral binning. But wouldn't it lose another unit of meter from the LOS integration, so the ultimate unit is ph/s/m2/sr?

cjperks7:
3) Going to just document it here while I'm at it. I'm observing the the brightness from Tofu is about an order of magnitude higher than the brightness values I got from the old workflow. I'll do a more quantitative analysis soon

Didou09:
I think it may come from the way the etendue is calculated, to be checked vs your analysis and vs the reverse ray-tracing I'm currently finishing

cjperks7:
But when brightness=True, it looks like multiplying by the etendue is skipped so I don't have to consider it right now

cjperks7:
I resolved my issue in (2) myself. I just hadn't defined units for my (R,Z) mesh so that my fault

Didou09:
> I resolved my issue in (2) myself. I just hadn't defined units for my (R,Z) mesh so that's my fault

Yes, Brightness = True essentially just does LOS-integration, and the units are picked up from the (R, Z) units

Didou09:
> But when brightness=True, it looks like multiplying by the etendue is skipped so I don't have to consider it right now

Yes, etendue is not used for Brightness = True

[...]

PR #751, opened 2023-04-27
Debug jumps in etendue for spectrometer concave crystals

dvezinet:
Main changes:
-----------------
Two minor bugs identified in the routines used to project an aperture (slit) on a concave crystal.
Concavity comes with numerical instability when the root-finding process comes close to the singular point that is the center of curvature.

Tofu implements a robust root-finding routine + interpolates reflected polygons to take out singular points.

**Bug 1:** the interpolation process was missing some corners of the polygons in some particular cases.

**Bug 2:** the root-finding algo wasn't robust enough for some particular pixels / configurations

Both bugs have been addressed in this PR, see issue #747 for examples

Several plotting / debugging sections have been added to the code to facilitate any later debugging in case another problem arises. They are commented out and left as such for later use.

Issues:
-------
* Fixes, in devel, issue #747

### DIFF
--- a/tofu/data/_class5_reflections_pts2pt.py
+++ b/tofu/data/_class5_reflections_pts2pt.py
@@ -155,7 +155,7 @@ def _get_pts2pt(
             return_x01=None,
             # number of k for interpolation
             nk=None,
-            debug=True,
+            debug=None,
             # timing
             dt=None,
         ):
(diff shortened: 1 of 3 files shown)


tofu/data/_class8_equivalent_apertures.py
```diff
@@ -237,10 +255,9 @@ def equivalent_apertures(
 
             elif ln[ii] < nmax:
                 ndif = nmax - ln[ii]
-                irand = np.random.random(ndif)
+                irand = np.random.randint(1, 9, ndif)/10.
                 irand = irand + np.random.randint(0, ln[ii]-1, ndif)
                 imax = np.sort(np.r_[np.arange(0, ln[ii]), irand])
-                imax = np.linspace(0, ln[ii]-1, nmax)
```
dvezinet, line 243: First bug was these 4 lines

tofu/data/_class5_reflections_pts2pt.py
```diff
@@ -247,21 +247,35 @@ def pts2pt(
...
+                    thr = abs(line[0] - line[-1]) * 0.05
+                    std = np.std(eq[ind] - line)
+                    if std > thr:
+                        continue
+                    roots = np.r_[-bb / aa]
+                    # roots = np.r_[
+                    #     kk[i0]
+                    #     - eq[i0]*(kk[i0+1] - kk[i0])
+                    #     / (eq[i0+1] - eq[i0])
+                    # ]
```
dvezinet, line 278: Second bug: more robust root-finding

cjperks7:
Just flipping plotting data through `diag.plot_diagnostic()` things look good except for `dlamb` and `res`. This is more so related to what's stated in Issue #748 than what's addressed here. Attached is a plot of dlamb from `diag.plot_diagnostic()`. Also attached is repeating the comparison plot given in Issue #748 to illustrate that the factor is still about off the same amount, but the shape changed from what you did here
Screenshot 2023-04-27 at 5 57 33 PM
Screenshot 2023-04-27 at 6 00 20 PM

cjperks7 (approved):

Didou09:
Well spotted (dlamb), as you mention, it's more related to issue #748 , will try to address that tomorrow morning

cjperks7:
Definitely changed some things up with the magnitude of `dlamb`. The prefactor to rescale it to just being the spacing in the wavelength mesh is now ~3.5 (previously ~4.3). 

Much less noisy though. The relative difference between the two meshes shown is now 1e-6%. So any difference in the first attached plot is really the subtraction I did for visualization.

The additional computational time isn't terrible. Might have slowed down "build diagnostic" section in tofu_sparc by say 50%-ish so not the end of the world.

Screenshot 2023-04-28 at 1 55 10 PM

PR #752, opened 2023-04-28
Correcting jumps in dlamb

Didou09:
Main changes:
-------------------
* After fixing bugs in reflected polygon root-finding and interpolation, @cjperks7  commented that while it had fixed the `etendue`, some instabilities remained in the computation of `dlamb`, see PR #751 for details
* After investigating, it comes not from a bug but simply from an unsufficient number of interpolation points for the polygon interpolation
* The number of interpolation points has been increased, resulting in better results at the cost of slower computation

Issues:
---------
* This partially fixes issue #748, as it addresses the stability aspect of `dlamb`, but does not yet answer why `dlamb` is generally larger than expected

Examples:
--------------

This is an example of a reflection of the slit as seen through the crystal by a pixel, without correction:

image

Same reflection after increasing the number of interpolation points:

image

The area is computed more accurately, resulting in a more accurate etendue, andmore points are tested for dlamb, resulting in better results too

After correction, dlamb is smoother:
image

### DIFF
--- a/tofu/data/_class5_reflections_pts2pt.py
+++ b/tofu/data/_class5_reflections_pts2pt.py
@@ -247,28 +247,27 @@ def _get_pts2pt(
                 iok = np.isfinite(eq)
                 kk = kk[iok]
                 eq = eq[iok]
-                
+
                 # find indices of sign changes
                 i0 = (eq[:-1] * eq[1:] < 0).nonzero()[0]
                 if len(i0) == 1:
                     # indices used for linear fit
                     ind = np.arange(max(0, i0-10), min(i0+10, kk.size))
-                    
+
                     # linear least square fit
                     xx_m = np.mean(kk[ind])
                     yy_m = np.mean(eq[ind])
                     xx2_m = np.mean(kk[ind]**2)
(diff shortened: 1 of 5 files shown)


cjperks7:
Definitely changed some things up with the magnitude of dlamb. The prefactor to rescale it to just being the spacing in the wavelength mesh is now ~3.5 (previously ~4.3).

Much less noisy though. The relative difference between the two meshes shown is now 1e-6%. So any difference in the first attached plot is really the subtraction I did for visualization.

The additional computational time isn't terrible. Might have slowed down "build diagnostic" section in tofu_sparc by say 50%-ish so not the end of the world.

Screenshot 2023-04-28 at 2 06 53 PM

cjperks7:
To clarify what I mean by that relative error bit. What I do in the tofu_sparc method (1) is fit the wavelength mesh out of `lamb=diag.get_diagnostic_lamb()` with a degree=2 polynomial so I can initiate a Tofu diagnostic on a coarse mesh to save time and interpolate it to whatever pixel mesh I want in the emissivity maps since it doesn't impact time here. So the output of `lamb-diag.get_diagnostic_lamb()` can be fit real good now.

Issue #756, opened 2023-05-04
Planar crystal: Discontinuity in the etendue


cjperks7:
Maybe this is due to me building things incorrectly, but I'm seeing a sudden drop in the etendue on the central vertical channel. This is with a (33,1) pixel mesh.

<img width="1390" alt="Screenshot 2023-05-04 at 2 28 12 PM" src="https://user-images.githubusercontent.com/71457911/236295815-446a7714-09c3-4370-b775-65873ef610db.png">

I've noticed when refining the vertical pixel mesh that when the number of pixels is even then that central hollowing widens and get populated with nan's. This is with a (33,4) pixel mesh:
<img width="1367" alt="Screenshot 2023-05-04 at 2 37 54 PM" src="https://user-images.githubusercontent.com/71457911/236297929-c08fb337-cfd0-4ca4-9953-073447944141.png">

Refining the horizontal mesh doesn't seem to impact this behavior

cjperks7:
Minimum working code. Same as attached in Issue #755 :
[make_tofu_spec.txt](https://github.com/ToFuProject/tofu/files/11400418/make_tofu_spec.txt)

The pixel mesh is defined by variables: `nx0, nx1`

PR #758, opened 2023-05-05
Issue756 flat crystal etendue

Didou09:
Main changes:
-------------------
* None, except a minor extra piece of copy-pasted code to help that kind of cases debugging 

Issues:
-----------
* Issue #756  was due to wrong input, no bug detected

Example:
------------

The outline of the crystal was provided such that the crystal was self-intersecting, resulting in reflected slit that was plummeting to 0 for central pixels

image

image
image

Indeed, the reflection of the slit on the crystal gradually disappears as we go from edge pixels to central pixels (index ii = 16)

image

A simple fix is to provide the outline in good order to avoid self-intersection of the polygon:
image

Which results in:

image

### DIFF
--- a/tofu/data/_class8_equivalent_apertures.py
+++ b/tofu/data/_class8_equivalent_apertures.py
@@ -219,6 +219,11 @@ def equivalent_apertures(
         x0.append(p0)
         x1.append(p1)
 
+        # --- DEBUG ---------
+        # if ii in [14, 15, 16, 17]:
+            # _debug_plot(pa0=p0, pa1=p1, ii=ii, tit='local coords')
+        # --------------------
+
     # -------------------------------------------
     # harmonize if necessary the initial polygons
     # -------------------------------------------
@@ -573,6 +578,8 @@ def _get_equivalent_aperture(
     nop_pre=None,
     lpoly_pre=None,
     ptsvect=None,
+    # debug
+    ii=None,
     **kwdargs,
 ):
 
@@ -590,6 +597,11 @@ def _get_equivalent_aperture(
             strict=False,
             return_x01=True,
         )[-2:]
+
+        # --- DEBUG ---------
+        # if ii in [14, 15, 16, 17]:
+            # _debug_plot(p_a=p_a, pa0=p0, pa1=p1, ii=ii, tit='local coords')
+        # --------------------
 
         # inside
         if np.all([p_a.isInside(xx, yy) for xx, yy in zip(p0, p1)]):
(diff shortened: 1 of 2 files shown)


cjperks7:
Thanks for catching my mistake! 

My one question is what's the quantity you're plotting in the bottom right of your last plot? Etendue? I wasn't able to reproduce it flipping through the options but may have missed something.

Didou09:
> Thanks for catching my mistake!
> 
> My one question is what's the quantity you're plotting in the bottom right of your last plot? Etendue? I wasn't able to reproduce it flipping through the options but may have missed something.

Yes, it is the etendue, are you sure you loaded the good branch ?

cjperks7:
Yeah, I think I am. I do `git checkout Issue756_FlatCrystalEtendue` and then run the minimum working example I passed to you yesterday with only this identified error. I'm getting an etendue of ~5e-14.

I'm interested in what's going on here because what you've plotted for etendue is more similar to what's come out for the von Hamos. I would expect to get numbers like that.

cjperks7:
After checking, running the von Hamos with the same detector pixel settings (33,1) is giving ~1.3e-13 so comparing against the flat's ~5e-14 is only about a factor 3 difference. Not sure how your ~2e-12 for flat pops up, but like we talked about more careful signal comparison coming up since you hashed out these coding bits

Didou09:
> After checking, running the von Hamos with the same detector pixel settings (33,1) is giving ~1.3e-13 so comparing against the flat's ~5e-14 is only about a factor 3 difference. Not sure how your ~2e-12 for flat pops up, but like we talked about more careful signal comparison coming up since you hashed out these coding bits

Didou09:
OK, I got it, this 1e-12 etendue corresponds to the etendue without the rocking curve effects.
It's because after creating the PR I had shofted branches to a branch on which I had been temporarily deactivating the rocking curve , made the plots from there ((forgetting I had swithed branches).

So 5e-14 is the correct answer

cjperks7:
> OK, I got it, this 1e-12 etendue corresponds to the etendue without the rocking curve effects. It's because after creating the PR I had shofted branches to a branch on which I had been temporarily deactivating the rocking curve , made the plots from there ((forgetting I had swithed branches).
> 
> So 5e-14 is the correct answer

Damn, yeah that checks out. I can recover this plot turning off the rocking curve

cjperks7:
Always helps to test our understanding of the numbers. Yeah, I think everything's good here

Issue #755, opened 2023-05-04
Planar crystal: provide extenthalf vs outline ?


dvezinet:

cjperks7:
I've noticed that when building a planar crystal spectrometer that when you use `diag.add_crystal()`, it requires the crsytal['dgeom'] to not have an 'extenthalf' entry:
<img width="839" alt="Screenshot 2023-05-04 at 2 18 43 PM" src="https://user-images.githubusercontent.com/71457911/236293671-4377a6ff-507c-4f50-8e7f-be3f98d1423f.png">

But, if you don't define an 'extenthalf' (so just 'outline_x0', 'outline_x1') then later in `diag.add_diagnostic()` you get an error basically requesting it:
<img width="839" alt="Screenshot 2023-05-04 at 2 20 42 PM" src="https://user-images.githubusercontent.com/71457911/236294082-57a5c6e9-b558-467b-ab87-8518fefebe97.png">

I've been able to work around this by manually adding 'extenthalf' after using `diag.add_crystal()`

cjperks7:
Minimum working code:
[make_tofu_spec.txt](https://github.com/ToFuProject/tofu/files/11400287/make_tofu_spec.txt)

PR #759, opened 2023-05-05
Issue755 flat extent vs outline

Didou09:
Main changes:
--------------------
* Apertures provided as outline only are now analyzed, if a rectangular shape is detected, extenthalf and curve_r are automatically computed

Issues:
---------
* Fixes, in devel, issue #755


cjperks7 (approved):
Works good

PR #765, opened 2023-05-23
Issue739 spectro reverse rt

Didou09:
Main changes:
-------------------
* Several improvement to vos computation for non-spectrometers
* [BUGFIX](https://github.com/ToFuProject/tofu/pull/765/commits/85e32382d53d1727cb5472aecd7f08bf6414688d): better interpolation of polygons in `equivalent_aperture()`
* [BUGFIX](https://github.com/ToFuProject/tofu/pull/765/commits/0653e639667e9dc6033e4396d2f0a40e638d6fd7): more rbust root finding in pathological cases for reflections on crystals
* [BUGFIX](https://github.com/ToFuProject/tofu/pull/765/commits/5230413b85ce743ca3738d332ac86da7e531f4d0): dlamb was overestimated by a factor ~2 from the LOS-approximation
* [BUGFIX](https://github.com/ToFuProject/tofu/pull/765/commits/3113fd46f270e7390c0fec5019c8ce87aaba5429): rocking curve now takes into account crystal to last aperture distance for LOS-approximation
* [BUGFIX](https://github.com/ToFuProject/tofu/pull/765/commits/a0a8b078f85f5a4d7f3a81d0220207ab6025ddf4): in some cases, reflection on a plane could lead to negative kk values that should be eliminated
* [Re-implemented: ](https://github.com/ToFuProject/tofu/pull/765/commits/2cda3aa5bc51146bc021d25d0c9618f12b6a625e)`compute_diagnostic_etendue_los(numerical=True)` for non-spectro
* `plot_diagnostic_vos()` implemented for spectro and non-spectro
* `compute_diagnostic_solidangle_from_plane()` operational for spectro and non-spectro
* `get_raytracing_from_pts()` implemented, only used for spectrometers
* `compute_diagnostic_vos(replace_poly=bool)` implemented to update vos polygons in cross-section
* `get_optics_reflections_ptsvect(isnorm=bool, fast=bool) `implemented for faster spectro reflections 
* some refactoring / moving routines around
* More debugging options (commented but available for later re-use)
* `tuto_real0.py` updated to easily create test spectrometers (flat crystal, cylindircal crystal and non-spectro equivalent)

On the whole, etendue, taking into account rocking curve and `dlamb`, now agree within **30-40 %** between the LOS approximation and a full VOS reverse ray-tracing (tested on a flat and on a cylindrical von hamos crystal)

Issues:
----------
Fixes, in devel, issues 
* #739
* #766 

Examples:
-------------

Using code from `tuto_real0.py`, we can quickly create 2 spectrometers, with respectively a flat crystal and a cylindrical (Von Hamos) crystal

```
In [1]: import tofu as tf

In [2]: test = tf.tests.tests09_tutorials.tuto_real0

In [3]: conf, coll = test.main()
```

Now, we use the method to generate a 2d plane perpendicular to a LOS of reference of the camera, sample that 2d plane to compute the locally at each point, and using both a LOS  approximation and a VOS / reverse ray-tracing approach.

For a non-spectrometer diagnostic (detector + apertures), it allows to check to visualize the 2d map of solid angle associated to a single channel and see from which resolution of the plane sampling the numerical etendue matches the analytical etendue.

```
In [4]: dout = coll.compute_diagnostic_solidangle_from_plane(key_diag='d0', key_cam='bb1', res=[0.01, 0.01], config=conf, margin_perp=[0.01,0.02], margin_par=0.5)
```
image

With a sufficient resolution we get a converged integral => etendues match
```
In [5]: dout = coll.compute_diagnostic_solidangle_from_plane(key_diag='d0', key_cam='bb1', res=[0.001, 0.001], config=conf, margin_perp=[0.01,0.02], margin_par=0.5)
```

image

Note that even after convergence (sufficient resolution), the VOS and LOS etendue only match at the reference channel, that's because the etendue is dependent on the direction of observation and the integral on the plane is only equal to the etendue for the LOS perpendicular to that plane.

For a spectrometer, the routine will compute and compare:

1. the solid angle (step 1 of the comparison)
     - subtended by each camera pixel (through the slit and via the reflection on the crystal)
     - purely geometric, not taking into account the rocking curve 
2. the solid angle convolved by the rocking curve  (step 2 of the comparison)
    - the LOS approcah uses an approximation based on the minimum between the typical angular width of the rocking curve and the angular width subtended by  pixel
3. the solid angle, convolved by the rocking curve, times the wavelength sampling step (dlamb), which gives the per-pixel quantity by which to multiply the LOS-integral at he reference wavelength to get the total photon flux per pixel.

This 3-steps comparison with increasing complexity was designed to facilitate debugging and understand where most of the differences come. They mostly come from how the 2 approaches model the rocking curve in the case of the flat crystal.

For the flat crystal:
```
In [4]: dout = coll.compute_diagnostic_solidangle_from_plane(key_diag='d03', key_cam='c0_camf', res=[0.0001, 0.005], config=conf, margin_perp=[0.01, 0.05], margin_par=13., n0=100, n1=100, res_lamb=0.02e-13)
		32507 / 32508 pts angmin dangmin: rock 2.36e-07 vs 1.65e-04 1/4 pixel
```

image

For the von hamos crystal, most of the difference comes from the ray-tracing (so probably from how the reflection of the slit is computed for the LOS-approach), later stages reduce the error.
```
In [5]: dout = coll.compute_diagnostic_solidangle_from_plane(key_diag='d02', key_cam='c0_cam', res=[0.0001, 0.005], config=conf, margin_perp=[0.01, 0.10], margin_par=13., n0=100, n1=100, res_lamb=0.02e-13
    ...: )
		18743 / 18748 pts angmin dangmin: rock 2.36e-07 vs 1.65e-04 1/4 pixel


```

image

Finally, the `compute_diagnostic_vos()` and `plot_diagnostic_vos()` methods samples a volumic mesh and computes the local solid angle (for non-spectro diags) convolved by the rocking curve (for spectro diags, it also stores extra quantities like the local cosine of the angle vs the toroidal direction).

For a non-spectrometer:
```
In [6]: dvos = coll.compute_diagnostic_vos('d0', key_mesh='m0', res_RZ=[0.01, 0.01], res_phi=0.001, visibility=False, store=False)

In [8]: dax = coll.plot_diagnostic_vos('d0', dvos=dvos, plot_config=conf)
```

image

For a spectrometer:
```
In [6]: dvos = coll.compute_diagnostic_vos('d03', key_mesh='m0', store=False, visibility=False, timing=False, verb=True, res_RZ=[0.05, 0.02], res_phi=0.0003, res_lamb=1e-15, res_rock_curve=25, debug=False,
   ...:  n0=10, n1=10)

Computing vos for diag 'd03':
Recommended res_lamb to ensure rocking curve overlap:
	- edge-edge: 	1.99e-14
	- MH-to-MH: 	3.27e-15
	- resolution: 	8.29e-16
	Provided: 	1.31e-15

	lamb.shape: (1248,)
	ang_rel.shape: (25,)
	iru.size: 31
	nRZ: 359

		30 / 31, 10 / 11, 407 / 408:  82 pts   	 dangmin: rock 3.94e-06 vs 1.72e-04 1/4 pixel


In [7]: dax = coll.plot_diagnostic_vos('d03', dvos=dvos)
```

image


Didou09:
@cjperks7 , @adriendaros , I'm putting you both as reviewers.
This is a big PR, so dont' feel like you have to go through everything, it's mostly to make sure you are informed of what is going on and you see what is possible / debugged / working.

Don't hesitate to try and reproduce some of these figures if you feel like digging a bit.

cjperks7:
Bit of a sanity check, you started this exercise to have an improved calculation with the VOS method over the LOS method but now you're quoting that the agree within 30-40%. Wouldn't you only expect that to be in with an asymptotically long and narrow spectrometer? How are you making that comparison?

Didou09:
The comparison is done on the total photon flux per LOS length, assuming an uniform emissivity of 1 ph/s/m3/sr/AA.

You get a photon flux per LOS length (ph/s/m) by mutipliying that unity emissivity by a factor computed as:

* LOS:  multiplying the etendue (taking into account the rocking curve) by dlamb (m2.sr x wavelrngth in m = sr.m2.AA)

* VOS: integrating over a plane perpendicular to the LOS (surface => m2) the local solid angle (sr) convolved by the rocking curve (computed as a sum of independent rays traced to the camera), integrated over a large sample of wavelength (dlamb) => sr.m2.AA

This is kind of a generilzation of the definition of a LOS and etendue

Issue #767, opened 2023-05-24
Compute geometry matrix from VOS for non-spectro diags


Didou09:
Motivations:
----------------
* Now that vos can be computed, it should be used for computing geometry matrices
* Let's start by non-spectro diags (easier)

PR #768, opened 2023-05-24
Issue767 geom matrix vos

Didou09:
Main changes:
-------------------
* The geometry marix for a non-spectrometer diag can now be computed using a VOS approach
* Help identify a minor bug in mesh outline computation, fixed in [bsplines PR 16](https://github.com/ToFuProject/bsplines2d/pull/16)

Issues:
--------
* Fixes, in devel,:
    * issue #767  

Examples:
--------------

Using `tuto_real0` we can create a set of diagnostics

```
In [11]: import tofu as tf

In [12]: test = tf.tests.tests09_tutorials.tuto_real0

In [13]: conf, coll = test.main()
```

Let's compute the geometry matrix for diag `d0` with a LOS approach:
 ```
In [14]: dmat = coll.add_geometry_matrix(key_diag='d0', key_bsplines='m0_bs1', method='los')
```

Let's compute a vos discretization (cf, PR #765) and then use it to compute a geometry matrix with a VOS approach:

```
In [16]: dvos = coll.compute_diagnostic_vos('d0', key_mesh='m0', res_RZ=[0.005, 0.005], res_phi=0.005, visibility=False, store=False, verb=True)

Computing vos for diag 'd0':
	cam 'bb1' pixel 10 / 10		npts in cross_section = 4538	(97716 total))	
```

And:
```
In [17]: dmat = coll.add_geometry_matrix(key_diag='d0', key_bsplines='m0_bs1', dvos=dvos, method='vos')
Geom matrix for diag 'd0' and bs 'm0_bs1':
	- 'gmat01_bb1': pixel 10 / 10s)	448 / 1980
	 bspline 187 / 198   (4538 pts)
```

Let's plot both geometry matrices:
```
In [18]: dax = coll.plot_geometry_matrix('gmat00', plot_config=conf, indchan=0, plot_mesh=False, res=0.005)     

In [19]: dax = coll.plot_geometry_matrix('gmat01', plot_config=conf, indchan=0, plot_mesh=False, res=0.005)
```

image

image

Note that the difference is not obvious in this case because the mesh is very coarse (10 cm resolution).
The difference will be more visible the finer the mesh gets and the lower the degree of the bsplines get too.

Below an example with a finer 3x3 cm2 mesh, larger fields of view (5 x 5 mm2 aperture) and deg=0 bsplines:

image

image


Didou09:
@cjperks7, @adriendaros   just for your information on these developments

VOS-computation of a geometry matrix for spectrometers is next

PR #907, opened 2024-02-21
Issue906 centroid + more robust inversion checks

dvezinet:
Main changes:
-----------------
* `save_to_step(include_centroid=bool)` implemented
* `add_inversion()` more robust vs zero-filled geometry matrices

Issues:
-------
Fixes, in devel, issue #906


dvezinet (commented):
good to go

Issue #930, opened 2024-05-28
Rounding error in unit tests for MacOS


dvezinet:
Rounding error on new MacOS compilers [in unit tests](https://github.com/ToFuProject/tofu/actions/runs/8972107290/job/25499118657?pr=928):

![image](https://github.com/ToFuProject/tofu/assets/112180133/baff4a81-0a7e-4e8f-90c7-d305288fdb0d)

PR #932, opened 2024-05-28
[#930] Specified tolerances for np.allclose(atol=1e-14)

dvezinet:
Main changes:
-----------------

`tests01_geom/test_01_GG.py`:
* `test16_dist_los_vpoly`: 
     - Specified tolerances for `np.allclose(atol=1e-12, rtol=1e-12)`, for clarity
     - Large discrepancy observed between MacOS and Linux on this particular unit test, for some specific cases

* `_distance_tools.pyx`: 
     - `c_sqrt()` was applied on a quantity that was `0` within machine accuracy but negative (`-2e-17`)
     - Resulting in a `NaN` that was propagating in certain special cases
     - Fix: apply `c_sqrt(c_abs())` for better robustness


dvezinet (commented):
done

